PR #13, opened 2015-03-29
Update nil check to support all nillable types.

dakiva:
http://golang.org/ref/spec#The_zero_value

There are 6 types for which the zero value is nil: pointers, functions, interfaces, slices, channels, and maps.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -279,7 +279,6 @@ func (g *Generator) Generate() error {
 		default:
 			g.printf("(%s) {\n", strings.Join(returs, ", "))
 		}
-
 		if len(types) > 0 {
 			g.printf("\tret := m.Called(%s)\n\n", strings.Join(names, ", "))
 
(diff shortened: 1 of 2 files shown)


evanphx:
As usual, great work!

spaowi:
What about named nillable types, for example:

    package main
    type B interface {
      Bar()
    }
    type A interface {
      Get() B
    }

Running `mockery -name A -print` on this does not yield a mock which checks for nil! The problem is that the B return type is parsed to an `*ast.Ident`, not an `*ast.InterfaceType` (and therefore isNillable returns false).

Perhaps you can find the type of the identifier to handle this case:

    func (g *Generator) isNillable(typ ast.Expr) bool {
            switch t := typ.(type) {
            case *ast.StarExpr, *ast.ArrayType, *ast.MapType, *ast.InterfaceType, *ast.FuncType, *ast.ChanType:
                    return true
            case *ast.Ident:
                    switch d := t.Obj.Decl.(type) {
                    case *ast.TypeSpec:
                            switch d.Type.(type) {
                            case *ast.StarExpr, *ast.ArrayType, *ast.MapType, *ast.InterfaceType, *ast.FuncType, *ast.ChanType:
                                    return true
                            }
                    }
            }
            return false
    }

mattyurka:
See https://github.com/vektra/mockery/issues/22

PR #16, opened 2015-04-07
Create unexported mocks from unexported interfaces

ernesto-jimenez:
When creating a mock with `-inpkg=true` for an unexported interface, the
generated mock should be unexported to avoid polluting the exported API
with an internal mock.

With this change, when generating a mock for `requester`, the generated
mock will be named `mockRequester` rather than `Mockrequester`.


evanphx:
Thanks!!

PR #17, opened 2015-04-08
Added info to README regarding main package mocking

warmans:
Added extra information regarding main namespace.


evanphx:
@warmans Thanks! Good to clarify that.

PR #21, opened 2015-05-16
Better file names for the output

### LINKED ISSUE #20: File names for -inpkg=true mocks should have the _test suffix
When generating an interface with `-inpkg=true` the generated file name is not suffixed with `_test`, which means the mock is always built.

The expected behaviour would be to add the `_test` suffix so the mock is only built for testing.

ernesto-jimenez:
Append `_test` to the -inpkg=true mocks so the file is only built for testing.

File names in lower case to be more aligned with Go's style.

Closes #20

### DIFF
--- a/cmd/mockery/mockery.go
+++ b/cmd/mockery/mockery.go
@@ -167,9 +167,9 @@ func genMock(iface *mockery.Interface) {
 		var path string
 
 		if *fIP {
-			path = filepath.Join(filepath.Dir(iface.Path), "mock_"+caseName+".go")
+			path = filepath.Join(filepath.Dir(iface.Path), filename(name, *fIP))
 		} else {
-			path = filepath.Join(*fOutput, caseName+".go")
+			path = filepath.Join(*fOutput, filename(name, *fIP))
 			os.MkdirAll(filepath.Dir(path), 0755)
 			pkg = filepath.Base(filepath.Dir(path))
 		}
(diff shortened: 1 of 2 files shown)


evanphx:
The reason I didn't add `_test.go` to the mock files is that doing so means that `go test blah_test.go` won't load the mocks. Perhaps we should add another flag to control adding `_test.go` so it doesn't confuse any current users.

ernesto-jimenez:
@evanphx adding a flag should work too, although it feels weird that the default `-inpkg` behaviour is to include the mock in the package's code rather than just the tests.

Out of curiosity, what's the use for `go test blah_test.go`?

I always use `go test` and `-run` when I need to filter which tests to run.

evanphx:
`go test blah_test.go` runs just the tests in that file.

ernesto-jimenez:
What flag would you like?

evanphx:
Perhaps `-testonly`?

ernesto-jimenez:
@evanphx I'm rebasing to master and just noticed tests in master are not passing. Are they OK on your side?

evanphx:
Ooops! Sorry about that! All green now, rebase away!

ernesto-jimenez:
@evanphx updated :)

cmd/mockery/mockery.go
```diff
@@ -217,3 +218,13 @@ func underscoreCaseName(caseName string) string {
 	rxp2 := regexp.MustCompile("([a-z0-9])([A-Z])")
 	return strings.ToLower(rxp2.ReplaceAllString(s1, "${1}_${2}"))
 }
+
+func filename(name string) string {
+	name = strings.ToLower(name)
```
evanphx: You don't need this anymore, caseName, the argument passed in, is already formatted properly.

ernesto-jimenez:
@evanphx removed the unnecessary line :)

evanphx:
Thanks!

ernesto-jimenez:
my pleasure :)

PR #25, opened 2015-06-26
Use the path to determine the mocks package

cam72cam:
When I specify -output=./mock -all
I expect the package to be named mock and not mocks.

This adds a override to the package name for that case.

### DIFF
--- a/mockery.go
+++ b/mockery.go
@@ -145,6 +145,7 @@ func genMock(iface *mockery.Interface) {
 
 	var out io.Writer
 
+	pkg := "mocks"
 	name := iface.Name
 
 	if *fPrint {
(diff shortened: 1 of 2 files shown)


mockery.go
```diff
@@ -9,7 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/vektra/mockery/mockery"
+	"github.com/treetopllc/mockery/mockery"
```
evanphx: This needs to be changed back to vektra

cam72cam:
@evanphx I pushed the import fix

evanphx:
@cam72cam Thanks!

PR #26, opened 2015-06-30
Return Value Provider Functions

ryanbrainard:
This adds optional support for mocks to use functions to provide the
return values at test time from arguments. This is a backward compatible
change and still supports standard return values. See [docs](https://github.com/ryanbrainard/mockery/blob/59ab6b5153d59cb24e5b9472a6bec1c093c3d382/README.md#return-value-provider-functions) for an example of how this can be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,16 @@ type Stringer struct {
 }
 
 func (m *Stringer) String() string {
- ret := m.Called()
+	ret := m.Called()
 
- r0 := ret.Get(0).(string)
+	var r0 string
+	if rf, ok := ret.Get(0).(func() string); ok {
+		r0 = rf()
+	} else {
+		r0 = ret.Get(0)
+	}
 
- return r0
+	return r0
 }
 ```
 
@@ -49,6 +54,32 @@ package to remove any unnecessary imports (as they'd result in compile errors).
 ### Types
 
 mockery should handle all types. If you find it does not, please report the issue.
+
+### Return Value Provider Functions
+
+If your tests need access to the arguments to calculate the return values,
+set the return value to a function that takes the method's arguments as its own
+arguments and returns the return value. For example, given this interface:
+
+```go
+package test
+
+type Proxy interface {
+  passthrough(s string) string
+}
+```
+
+The argument can be passed through as the return value:
+
+```
+Mock.On("passthrough").Return(func(s string) string) {
+    return s
+})
+```
+
+Note, this approach should be used judiciously, as return values should generally 
+not depend on arguments in mocks; however, this approach can be helpful for 
+situations like passthroughs or other test-only calculations.
 
 ### All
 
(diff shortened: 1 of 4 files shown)


evanphx:
My main concern with this change is that it breaks any interface that would be legitimately returning a function since you overload returning a function to mean that it should be called to provide the value rather than the function boing the value itself.

Rather than it being a bare function that is tested for, if it were a pointer to a struct, where the struct is defined by the mock package and the struct contains the function to be called, then at least we can say in the docs "If the mock detects the value is a *FuncValue object, it will call it provide the return value".

ryanbrainard:
This should still work with methods that return functions. If the method returns a function, the return value provider would be function of a function (and the plain value would just be a function). See the [test](https://github.com/ryanbrainard/mockery/blob/func-ret/mockery/generator_test.go#L482-L495) of this [interface](https://github.com/ryanbrainard/mockery/blob/func-ret/mockery/fixtures/func_type.go#L6) for an example.

ryanbrainard:
Just to be sure, I tried using the `Baz` mock in the example/test above in an actual test using mocks, and it worked as advertised:

```
func (n *TestSuite) TestGenFunc(c *C) {
	n.activities.Mock.On("Baz", mock.Anything).Return(func(x string) string {
		return x + "c"
	}, nil)

	bazFunc, _ := n.activities.Baz("a")
	c.Assert(bazFunc("b"), Equals, "bc")
}

func (n *TestSuite) TestGenFuncFunc(c *C) {
	n.activities.Mock.On("Baz", mock.Anything).Return(func(w string) func(string) string {
		{
			return func(x string) string {
				return w + x + "c"
			}
		}
	}, nil)

	bazFunc, _ := n.activities.Baz("a")
	c.Assert(bazFunc("b"), Equals, "abc")
}
```

evanphx:
@ryanbrainard Yes, but you're forcing anyone that does want to return a function to use your new function returning behavior to do so. This breaks compatibility unnecessarily, imho. Using a specific type like I indicated doesn't break compatibility and introduces only a minor change when registered the function to be called.

ryanbrainard:
@evanphx As shown in `TestGenFunc` above where the method interface is `Baz(string) (func(string) string, error)`, mocks can still return a function (i.e. `func(string) string`) without using this new behavior. `TestGenFunc` works both on master and this branch without any changes. `TestGenFuncFunc` is an example of using this new behavior where the mock returns a function of the return value (i.e. `func(string) func(string) string`). Whatever the return value is, the provider is a function returning that can value, so will have a different type than the value itself (even if it is itself a function), so should be backward compatible. Please let me know if I'm missing something.

ryanbrainard:
@evanphx Just wanted to check back in on this. Can you please let me know if there is in fact a backward incompatibility on this?

evanphx:
I considered this a bit more this morning and it's fine. The way window of confusion about returned value types is so small I doubt anyone will hit it.

PR #29, opened 2015-07-26
Add -recursive option

ryanbrainard:
This unifies `-name` and `-all` to both not be recurse by default, but provides the option to do so. 

Previously ,`-name` was never recursed, while `-all` was always recursed. This change is helpful for  finding either a specific interface name in a multi-package project or limiting the depth of `-all`. This also simplifies the code a bit, as `checkDir` and `mockFor` are no longer needed.

This is a slight breaking change. Users depending on the previously recursive behavior of `-all` will need to add use `-recursive` to traverse into subdirectories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,9 +85,14 @@ situations like passthroughs or other test-only calculations.
 
 It's common for a big package to have a lot of interfaces, so mockery provides `-all`.
 This option will tell mockery to scan all files under the directory named by `-dir` ("." by default)
-and generates mocks for any interfaces it finds.
+and generates mocks for any interfaces it finds. To include subdirectories, also use `-recursive`.
 
 `-all` was designed to be able to be used automatically in the background if required.
+
+### Recursive
+
+Use the `-recursive` option to search subdirectories for the interface(s).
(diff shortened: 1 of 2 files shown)


evanphx:
Yeah, the breaking change here is going to be confusing for users. We probably shouldn't change `-all` being recursive. Probably need to add a different, non-recursive version, perhaps `-dir .` ?

ryanbrainard:
I don't think `-dir .` would be good because that would not allow users to be non-recursive in a non-`.` directory. 

I'm thinking a solution might be to make `-name` accept a regex. This would allow `-name .*` to be a non-recursive form of `-all`, and get a bonus feature of filtering the name with a regex.

The current implementation short-circuits the search when using `-name` after the interface is found, and we could keep that behavior when a non-regex is provided (i.e. an exact, plain-text match), but if a regex is provided, it would search the whole directory (and walk the subdirectories if `-recursive` is provided). 

I can try this out and see how if it works out.

ryanbrainard:
@evanphx I just updated this as explained above.

evanphx:
Looks great!

PR #36, opened 2015-08-03
Generate MockOn_* and Name_* methods

ryanbrainard:
To avoid writing the `methodName` param in `testify/mock` as a `string`, this adds generated helper methods:

 - `MockOn_*`: takes `interface{}` as params and returns a mock for the method
 - `MockOnTyped_*`: takes typed params and returns a mock for the method
 - `MockOnAny_*`: returns a mock for the method with `mock.Anything` for all params
 - `Name_*`: returns the method name itself (used for other methods that need `methodName` besides `Mock.On`)

Examples of expansion:

```
activities.MockOn_Foo(1, 2):
activities.Mock.On("Foo", 1, 2)

activities.MockOnTyped_Foo(1, 2):
activities.Mock.On("Foo", 1, 2))

activities.MockOnAny_Foo():
activities.Mock.On("Foo", mock.Anything, mock.Anything)
```

I believe this is preferable to obtaining the method name by reflection because its less repetitive, results in shorter mocks, can be typed, and avoids common pattern of `mock.Anything` for all params. It of course does require code gen, but if using mockery, projects would be doing this anyway.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -49,6 +49,12 @@ func (g *Generator) GenerateIPPrologue() {
 	g.printf("\n")
 }
 
+func (g *Generator) generateMockOn(variant string, fname string, builderParams []string, onParams []string) {
+	g.printf("func (_m *%s) MockOn%s_%s(%s) *mock.Mock {\n", g.mockName(), variant, fname, strings.Join(builderParams, ", "))
+	g.printf("\treturn _m.Mock.On(%s)\n", strings.Join(append([]string{"\"" + fname + "\""}, onParams...), ", "))
+	g.printf("}\n")
+}
+
 func (g *Generator) mockName() string {
 	if g.ip {
 		if ast.IsExported(g.iface.Name) {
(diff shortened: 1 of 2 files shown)


evanphx:
I'm not sure I like littering the mocks with so many methods. I'm concerned about colliding with methods that the mock implements slightly.

Rather than have so many different variants, can we just generate the most useful one? I think that would be the `MockOnTyped_Name` version. If we just generated it as `MockName` I think that's best. The other variants can be done by using `On` directly.

ryanbrainard:
The problem with only generating `MockOnTyped_Name` is that it will not work with `mock.Anything`. `MockOn_Name` is the most generic, but then it's not typed. `MockOnAny_Name` could be dropped, but personally I've found that I use it the most often in projects.

What about adding flags to opt in to the generation of these? e.g. `-mockon`, `-mockontyped`, `-mockonany`

evanphx:
That's a good idea, add flags to generate them. The question is what should the default be. I'm thinking the typed version is probably the most useful one from my perspective, so should that be `Mock<Name>`?

ryanbrainard:
Are you saying to rename `MockOnTyped_<Name>` to `MockOn_<Name>` and then rename what was `MockOn_<Name>` to something like `MockOnUntyped_<Any>`? The reason I named `MockOn_<Name>` the way it is that is matches closest to what `mock.MockOn` does with the untyped params.

evanphx:
Yeah, that's what I'm saying, but I'm saying also go a step further and add `Mock<Name>` (no `On_`) by default as well.

ryanbrainard:
Ok, I can do that, but then what would the other methods be called? It would be strange to have the "untyped" and "any" methods have a different naming convention.

evanphx:
@ryanbrainard That's true. Maybe switch the order around? Have `Mock<Name>` (with typed args), `Mock<Name>_Any` (for anys) and `Mock<Name>_Untyped` (for interface{})?

evanphx:
@ryanbrainard Any update on this?

gaffo:
It wouldn't be that hard to have a series of outputters that you could optionally turn on in the code with flags which let people additional features. It also wouldn't be too hard to send the json rep of the Interface object via stdin to another, user specified, application which could output whatever it wanted to stdout to be included by mockery into the interface file which would be even more flexible for growth.

PR #42, opened 2015-09-21
Add support for multiple named return variables of the same type

dwlnetnl:
Fix bug where the mock method signature for a function with multiple
named return variables of the same type is not generated correctly.

### DIFF
--- a/mockery/fixtures/requester_ret_elided.go
+++ b/mockery/fixtures/requester_ret_elided.go
@@ -1,0 +1,5 @@
+package test
+
+type RequesterReturnElided interface {
+	Get(path string) (a, b int, err error)
+}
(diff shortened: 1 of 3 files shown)


evanphx:
Thanks!

dwlnetnl:
When I've try to use this code on my Interface it gives a runtime error: index out of range. I'm currently looking into that.

evanphx:
No problem.

dwlnetnl:
It panics on ```g.isNillable(ftype.Results.List[idx].Type)``` in ```Generator.Generate``` when there are more ```returnTypes``` than elements in ```List```.

dwlnetnl:
See #43.

PR #49, opened 2016-01-14
Fix unit test caused by mockery several changes

isArbit:
Fix unit test caused by several commits in package mockery.

```
PASS
ok  	{my_path}/mockery/mockery	0.027s
```


evanphx:
@isArbit Thank you so much!

isArbit:
@evanphx No problem!

PR #59, opened 2016-04-06
Fix for #58 testonly flag broken

### LINKED ISSUE #58: Generating a mock with -testonly without setting -inpkg does not append _test to the file
n/t

gaffo:
fixes #58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+all: fmt test install
+
+fmt:
+	go fmt ./...
+
+test:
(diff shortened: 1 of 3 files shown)


evanphx:
Thanks!

gaffo:
Ugh there's some bug introduced into the recursion of the directories... Please revert this, I'm adding some testing around this part too so it won't happen again. It's not finding all mocks correctly.

Issue #51, opened 2016-02-01
Mockery should allow me to generate multiple interfaces in one line


gaffo:
I've got a bunch of interfaces I'd like to generate but mockery only allows me to specify one per invocation. There's a ton of other interfaces I don't want to generate in the same code paths so I can't do the ... form. Let me just specify multiple paths on the same invocation.

evanphx:
You can specify `-all`. But I'm also fine having `-name` take multiple arguments and could even take a regex to match. Would that work?

gaffo:
Yes that would be great

gaffo:
So are you thinking multiple -names? or comma seperated or regexp?

gaffo:
I've got approval to work on this at work, not guaranteed to do it :)

gaffo:
Can you re-open so I can work towards it?

PR #62, opened 2016-04-06
Seperate config parsing from main

gaffo:
Moving config parsing out to it's own function to make it easier to add complicated command line flags like repeating ones in #51


gaffo:
Rebased off of master and tested

gaffo:
Rescinding this in favor of 64

PR #63, opened 2016-04-06
Move output from main

gaffo:
This follows on from #62.

Moves the output stream creation into it's own type / interface and simplifies some of the creation logic into types. This will lead to being able to moving out a "walker" and a "generator". The interface also means we can mock test these types independent of the filesystem.

### DIFF
--- a/cmd/mockery/mockery.go
+++ b/cmd/mockery/mockery.go
@@ -29,20 +29,7 @@ type Config struct {
 }
 
 func main() {
-	config := Config{}
-
-	flag.StringVar(&config.fName, "name", "", "name or matching regular expression of interface to generate mock for")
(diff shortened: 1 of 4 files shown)


evanphx:
This looks good, but could you please try to batch your work up into one PR? I think I've merged 3 of yours today and it's becoming a bit hard to follow.

evanphx:
For instance, this one and #62. I don't see a reason to merge a PR that moves something from point A to B, then the next one that moves it from B to C.

gaffo:
Okay, good to know. IU was trying to get testability into the base as I worked towards actually fixing #51. I'll rescind 63 and try and glob them together.

gaffo:
Merged the 2 PRs.

gaffo:
Closing and folding into #64

PR #64, opened 2016-04-07
Pull all generation logic from mockery.go, which now only deals with command line args and setting up the generate.

gaffo:
This follows on  #63 and moves the walker and the generator out to its own types. Also adds some unit tests around the walking. This is in support of testing and augmenting the walking logic.

### DIFF
--- a/mockery/outputter_test.go
+++ b/mockery/outputter_test.go
@@ -1,0 +1,36 @@
+package mockery
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFilenameBare(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: false, TestOnly: false}
+	assert.Equal(t, "name.go", out.filename("name"))
+}
+
+func TestFilenameMockOnly(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: true, TestOnly: false}
+	assert.Equal(t, "mock_name.go", out.filename("name"))
+}
+
+func TestFilenameMockTest(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: true, TestOnly: true}
+	assert.Equal(t, "mock_name_test.go", out.filename("name"))
+}
+
+func TestFilenameTest(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: false, TestOnly: true}
+	assert.Equal(t, "name_test.go", out.filename("name"))
+}
+
+func TestUnderscoreCaseName(t *testing.T) {
+	assert.Equal(t, "notify_event", (&FileOutputStreamProvider{}).underscoreCaseName("NotifyEvent"))
+	assert.Equal(t, "repository", (&FileOutputStreamProvider{}).underscoreCaseName("Repository"))
+	assert.Equal(t, "http_server", (&FileOutputStreamProvider{}).underscoreCaseName("HTTPServer"))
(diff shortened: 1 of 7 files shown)


evanphx:
I'm getting lost in your PRs. You say this is a follow on to #63 but it seems to define the same things? Please clarify.

gaffo:
So these are basically chains of pull requests where each stacks on the other. Github doesn't do a good job of letting you diff one PR off of another. I'm trying to do them as atomic groups of commits without going too far down any one road. I've commited to repos before where I added things like make or integration test and they flipped out and so I had to go back and yank all that stuff. So I was taking a cautious iterative refactor approach.

These series of commits are really trying to get all of the logic out of mockery.go and have it deal solely with parsing a set of command line arguments and then kicking off another object. I feel this is way more reusable and much better for testing and maintaining command line apps.

Also yesterday morning you were merging right along with me so I went with a smaller approach. I'll favor this one over #63 as well as it's everything except the actual fix for #51 which I'm still iterating on. It does however do all of the refactors I think ar needed aroudn mockery.go which is to get all the logic out of it and to make the rest of that stuff testable.

On another note... How do you feel about using and generating actual mockery mocks in the source itself. There's a bit of a chicken and egg problem there but I think there are a few integration tests that would be way better in GO directly and would make the tool way safer to develop on. Coding integration tests in bash/make isn't fun :) I was considering just checking them in.

mockery/outputter_test.go
```diff
@@ -0,0 +1,36 @@
...
+	assert.Equal(t, "mock_name_test.go", out.filename("name"))
+}
+
+func TestFilenameTest(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: false, TestOnly: true}
+	assert.Equal(t, "name_test.go", out.filename("name"))
+}
+
+func TestUnderscoreCaseName(t *testing.T) {
+	assert.Equal(t, "notify_event", (&FileOutputStreamProvider{}).underscoreCaseName("NotifyEvent"))
```
gaffo: I know this is ugly, basically testing a private method. I think it'll be at a point soon where the function will be testable without the struct again. Maybe with a smaller interface.

mockery/outputter_test.go
```diff
@@ -0,0 +1,36 @@
+package mockery
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFilenameBare(t *testing.T) {
+	out := FileOutputStreamProvider{InPackage: false, TestOnly: false}
```
gaffo: I know testing a private object function is ugly but this is a move from testing a private function. I think it's a good intermediary step and not that far from being able to just test the object that is evolving.

evanphx:
Thanks for the work! I was merging along with you yesterday because I had no idea you had 5 more iterative PRs in the works. :grin: 

I prefer a PR to encapsulate a set of features of changes. I don't like PRs that are expected to build on other PRs. Those are difficult to understand and make the git history difficult to understand.

Do you feel that this PR in it's current state encapsulates what you're trying to accomplish or are there more coming that refine this?

gaffo:
So I'm debugging the actual feature, allow multiple names/features. I'm close on it but I'm taking a break to do some work on another project. So I think this commit gets mockery.go in a much better and more testable place but it doesn't actually implement any new features.

So it's up to you if you want to merge this as is or wait till I get #51 done.

evanphx:
@gaffo That's fine. Sounds good to merge. Thanks for the great work!

PR #72, opened 2016-04-08
Implement the documented Return functions behavior

abourget:
```
mymock.On("functionName").Return(func(a, b string) (string, string) {
    return fmt.Sprintf("%s + %s", a, b), "hello"
})
```

It is documented like so in the README.md, but doesn't work with 2 return values properly.  Now it does.

You'll need to regenerate your mock.

Tests pass... it seems mock generation is broken for other reasons.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -308,7 +308,11 @@ func (g *Generator) Generate() error {
 		params := genList(ftype.Params(), ftype.Variadic())
 		returns := genList(ftype.Results(), false)
 
-		g.printf("// %s provides a mock function with given fields: %s\n", fname, strings.Join(params.Names, ", "))
+		namesPrefix := " "
+		if len(params.Names) == 0 {
+			namesPrefix = ""
+		}
+		g.printf("// %s provides a mock function with given fields:%s%s\n", fname, namesPrefix, strings.Join(params.Names, ", "))
 		g.printf("func (_m *%s) %s(%s) ", g.mockName(), fname, strings.Join(params.Params, ", "))
 
 		switch len(returns.Types) {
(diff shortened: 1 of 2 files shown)


evanphx:
This effectively breaks all current users of that feature so I can't merge it. The semantics of your version are a bit odd since it allows a single return value to control both of the return values. I see where you're going with your example, but again it breaks all existing users of this function, so I simply can't merge it.

abourget:
wait a second though.. I can _not_ believe a single user would use such a convoluted call:

```
type Nice interface {
   Get(string) (*Profile, error)
}

mock.On("thisfunc").Return(func(a string) *Profile {
    if a == "hello" {
        return &Profile{}
    }
}, func(a string) error {
    if a == "hello" {
        return nil
    }
})
```

this is what current users would have to do to use this previous implementation. 

My bet is that users only use that feature with a single return param, and in this case, this PR will not break for them.

Please leave this open for discussion at least :)

evanphx:
There is simply no way to know, honestly. If you believe your version is the proper one, modify the PR to control the behavior behind a flag. In time, we can migrate the flag to default on.

abourget:
Hmm.. I don't know what to say.. In my eyes, the previous behavior is a very broken one.. I don't see a need to keep it around.. plus, we'd need to massively duplicate tests.

I mean.. people generate their mocks once.. it's not like we'd be breaking backwards compatibility so badly, is it ?

I'd obviously prefer not to keep a fork around..

Should we give this a couple of days to think about it ?

evanphx:
I understand that you think the current behavior has no merit, but regardless of that it is existing behavior that users depend on.

It does break backwards compatibility because anyone that upgrades has has existing mocks that they update (which is pretty common) will suddenly find that the tests are blowing up because functions with the wrong signatures are now in the return slots.

I have a duty to my users not to break existing behaviors silently as you're proposing.

I believe I've given you a satisfactory answer, that I'm happy to include the behavior behind a flag. It will not massively duplicate the tests as all that is required is one additional test to verify your new behavior.

evanphx:
I opt'd against this API.

Issue #74, opened 2016-04-14
Errors are swallowed up, making debugging difficult.


abourget:
in `doWalk` in `walker.go`, I tweaked lines 57...:

```go
		err = p.Parse(path)
		fmt.Println("Parse", path)
		if err != nil {
			fmt.Println("Error:", err)
			continue
		}
```

And running it I now read:

```
Parse store.go
Error: open .#main.go: no such file or directory
```

otherwise, it would say nothing, except it didn't find it..

What would be the best way to resurface errors ?

abourget:
Also, with the current revision, the generated mock files are empty ! Have you seen such a behavior ?

abourget:
in `ouputter.go`, line `53`, there is a `defer f.Close()` .. how can the caller write to a closed file ?

am I missing something ?

IvanMalison:
Yeah this makes development much more difficult.

evanphx:
#75 should have fixed this.

PR #75, opened 2016-04-14
Fixes empty mock files, and adds small error logging

abourget:
Addresses https://github.com/vektra/mockery/issues/74


abourget:
I've added a fix to also solve a broken `-inpkg` behavior.. when generating for a `main` package, all structs were prefixed with `main.`.

PR #76, opened 2016-04-19
use first component of $GOPATH for relative path

cstrahan:
On certain setups it's not uncommon to have multiple entries in $GOPATH.
For example, NixOS provides packaging for Go apps and libraries, and
those libraries get automagically added to $GOPATH (I'm handwaving a
bit, as this commit is too small to talk about the intricacies -- and
wonders -- of Nix & NixOS).
So, let's use specifically the first component of $GOPATH.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -72,7 +72,7 @@ func (g *Generator) mockName() string {
 func (g *Generator) GeneratePrologue(pkg string) {
 	g.printf("package %v\n\n", pkg)
 
-	goPath := os.Getenv("GOPATH")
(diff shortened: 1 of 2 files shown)


mockery/generator.go
```diff
@@ -72,7 +72,7 @@ func (g *Generator) mockName() string {
 func (g *Generator) GeneratePrologue(pkg string) {
 	g.printf("package %v\n\n", pkg)
 
-	goPath := os.Getenv("GOPATH")
+	goPath := strings.SplitN(os.Getenv("GOPATH"), string(os.PathListSeparator), 2)[0]
```
IvanMalison: Use filepath.SplitList

Issue #81, opened 2016-05-21
"could not import"


atombender:
```shell
$ mockery -name=Client -case=underscore
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Error parsing file:  client_test.go:14:2: could not import github.com/stretchr/testify/assert (can't find import: github.com/t11e/go-pebbleclient/vendor/github.com/stretchr/testify/assert)
Unable to find Client in any go files under this path
exit status 1

$ ls vendor/github.com/stretchr/testify/assert   
assertion_forward.go	   assertions.go       doc.go	  forward_assertions.go       http_assertions.go
assertion_forward.go.tmpl  assertions_test.go  errors.go  forward_assertions_test.go  http_assertions_test.go

$ pwd
/Users/alex/.go/src/github.com/t11e/go-pebbleclient

$ echo $GOPATH
/Users/alex/.go
```

Works with every other project I have, just not exactly this one. The project compiles and tests fine, there's nothing unusual about it. Same versions of Mockery and Testify, too.

deadok22:
Same problem here. Reverting to ```47a058062601e235731714f89d7ec95c1ec47a5e``` worked for me.

[...]

Issue #80, opened 2016-05-21
-inpkg produces wrong qualified symbols in non-main package


atombender:
If I use `-inpkg` and my package name is `things`, and my interface is `Thing`, then it generates a file in the correct package, but referencing the original interface as `things.Thing`:

```
package things

import "github.com/stretchr/testify/mock"

// MockThing is an autogenerated mock type for the Thing type
type MockThing struct {
	mock.Mock
}

// Get provides a mock function with given fields: uid, options
func (_m *Client) Get(uid string, options things.GetOptions) (*things.Item, error) {
	ret := _m.Called(uid, options)
}
[...]
```

But Go doesn't allow you to qualify symbols in a package with its own package name (`things.Item`), and running this through `goimports` will of course cause it to import itself. 

The correct behaviour is to not prefix the symbols that are in the same package.

maciejmrowiec:
+1

Having same issue

IvanMalison:
@atombender you may want to check to see if my pull request #93 has resolved this issue. or if joshk0s pr did so.

Issue #73, opened 2016-04-11
Broken file loading


dansimau:
bb38420 breaks generating mocks from within a package which also imports those mocks.

Consider the following `foo` package:

**`foo.go`:**
```go
package foo

type FooInterface interface {
	FooFunc()
}
```

**`foo_test.go`:**
```go
package foo

import "mocks" // Tests for foo depend on the mock

func Test() {
	f := &mocks.FooInterface{}
	f.FooFunc()
}
```

Previously, (at b08b323) the mock would be generated:

```
$ mockery --name=FooInterface
Generating mock for: FooInterface
```

However, after commit bb38420, the mocks are no longer generated (presumably the import of `foo_test.go` fails, because the mock hasn't been generated yet):

```
$ mockery --name=FooInterface
Unable to find FooInterface in any go files under this path
```

gist containing the files are here:
https://gist.github.com/dansimau/743bb5ea1e94355402d2e4e88b1d3f68

dskarataev:
guys, somehow the commit was not integrated into mockery master branch https://github.com/arbortech/mockery/commit/011419f93a9d71df8e3be78e345cd71c899b66c2

without this commit the problem still exists!

@joshk0, @evanphx 

please integrate it, thanks

dskarataev:
also, I use mockery through go code generation tool:

```
//go:generate mockery -name IAnswerConverter -output mocks
```

and I have this error while generating the mock in the "converter" package because I import this mock in the "converter_test" package under the same directory:

```
question_converter_test.go:6:2: could not import sample_api/internal/handler/converter/mocks (cannot find package "sample_api/internal/handler/converter/mocks" in any of:
	/Users/user/go/src/ask_api/vendor/sample_api/internal/handler/converter/mocks (vendor tree)
	/Users/user/go/src/vendor/sample_api/internal/handler/converter/mocks
	/Users/user/.gobrew/versions/1.7.1/src/sample_api/internal/handler/converter/mocks (from $GOROOT)
	/Users/user/go/src/sample_api/internal/handler/converter/mocks (from $GOPATH))
Error parsing file:  couldn't load packages due to errors: /Users/user/go/src/sample_api/internal/handler/converter/answer_converter.go
question_converter_test.go:6:2: could not import sample_api/internal/handler/converter/mocks (cannot find package "sample_api/internal/handler/converter/mocks" in any of:
	/Users/user/go/src/ask_api/vendor/sample_api/internal/handler/converter/mocks (vendor tree)
[...]

PR #87, opened 2016-06-05
Many mockery fixes

joshk0:
- Fix qualified symbol name for InPackage mode (Issue #80)
- Ignore _test.go files when building mocks. (Issue #73)
- Use golang.org/x/tools/go/loader instead of go/parser. (Issue #81)

### DIFF
--- a/mockery/parse.go
+++ b/mockery/parse.go
@@ -3,11 +3,11 @@ package mockery
 import (
 	"go/ast"
 	"go/importer"
-	"go/parser"
-	"go/token"
 	"go/types"
+	"golang.org/x/tools/go/loader"
 	"io/ioutil"
 	"path/filepath"
+	"strings"
 )
 
 type Parser struct {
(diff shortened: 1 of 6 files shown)


mockery/parse.go
```diff
@@ -3,11 +3,11 @@ package mockery
 import (
 	"go/ast"
 	"go/importer"
-	"go/parser"
-	"go/token"
 	"go/types"
+	"golang.org/x/tools/go/loader"
 	"io/ioutil"
 	"path/filepath"
+	"strings"
```
fastest963: This isn't used anywhere

fastest963:
It looks like the switch to using `loader` instead of `parser` fixed my issues that I had in #81 but when deleting the mocks folder and running mockery, it still complains about `*_test.go` files importing `mocks`:
```$ mockery -name='Route53'
r53_test.go:14:17: undeclared name: mocks
Error parsing file:  couldn't load packages due to errors: /opt/gopath/src/bitbucket.org/levenlabs/vigilante/r53/r53.go
r53_test.go:14:17: undeclared name: mocks
```

joshk0:
Sorry, the strings was for doing a strings.HasSuffix(fi.File(), "_test.go") -- somehow it got lost when i chunked this into several commits. Will file a follow up PR for that.

Issue #47, opened 2016-01-04
Imports of paths with the same suffix in Interface file are not handled correctly


IvanMalison:
mockery does not seems to detect this case and provide an alias for the second package with the same name, which results in a 'redeclared as imported package name' error

evanphx:
Could you give me a more concrete example?

IvanMalison:
> Could you give me a more concrete example?

Sure

``` go
package example

import my_mockery "github.com/IvanMalison/mockery"
import "github.com/vektra/mockery/mockery/"

// Example is an example
type Example interface {
	K() my_mockery.A
	J() mockery.Cleanup
}
```
yields
``` go
package mocks

import "github.com/stretchr/testify/mock"

import "github.com/vektra/mockery/mockery/"

// Example is an autogenerated mock type for the Example type
type Example struct {
	mock.Mock
}

// J provides a mock function with given fields:
func (_m *Example) J() mockery.Cleanup {
	ret := _m.Called()

	var r0 mockery.Cleanup
	if rf, ok := ret.Get(0).(func() mockery.Cleanup); ok {
		r0 = rf()
	} else {
		if ret.Get(0) != nil {
			r0 = ret.Get(0).(mockery.Cleanup)
		}
	}

	return r0
}

// K provides a mock function with given fields:
func (_m *Example) K() mockery.A {
	ret := _m.Called()

	var r0 mockery.A
	if rf, ok := ret.Get(0).(func() mockery.A); ok {
		r0 = rf()
	} else {
		r0 = ret.Get(0).(mockery.A)
	}

	return r0
}
```

PR #92, opened 2016-06-25
Handle explicit import naming in mockery generator

IvanMalison:
This required two separate fixes:

The first of these was to lookup package aliases when making a call to
renderType so that references actually go through the type alias. The
second was to lookup the actual file within a package that contains the
interface definition so package imports are properly added in
GeneratePrologue.

Fixes: #47


IvanMalison:
Obviated by #92

PR #99, opened 2016-07-22
Add travis.yml

kevinburke:
The tests take a while to run, so it makes sense to have an external builder
run them on PR's, and email you the results.


kevinburke:
Here's what this looks like: https://travis-ci.org/kevinburke/mockery/builds/146720772

PR #93, opened 2016-06-27
Refactor import generation to be much more robust

IvanMalison:
Fixes:  #89 #47 #94 #77 #96

PR #102, opened 2016-09-09
Skip *_test.go files

fumoboy007:
This fixes a problem I was seeing relating to missing imports when using `-all`.

Also, there should not be interfaces in test files, so we do not need to parse them.

### DIFF
--- a/mockery/walker.go
+++ b/mockery/walker.go
@@ -48,7 +48,7 @@ func (this *Walker) doWalk(dir string, visitor WalkerVisitor) (generated bool) {
 			continue
 		}
 
-		if !strings.HasSuffix(path, ".go") {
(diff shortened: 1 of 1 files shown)


fumoboy007:
Also, this fixes the niggling problem of double output at the command line.

For example:
```
Generating mock for: MyInterface
Generating mock for: MyInterface
```

IvanMalison:
I'm not sure that I agree that this should have been merged.

fumoboy007:
Why?

IvanMalison:
Because the missing import problem can be solved in other ways, and I have actually written tests in the past where I created a mock for an interface defined in a test file.

IvanMalison:
I think that #93 might have fixed this.

fumoboy007:
OK I concur.

PR #105, opened 2016-10-26
Ignore _test.go files in parser

tehleach:
Following up on change in #102. #102 fixed **re**generating mocks with `_test` package in the same directory as the code when mocks exist, but if I then deleted all the mocks and tried to generate them "fresh", the mock generation failed because parser was still parsing all `.go` files in the directory **including** `_test.go` files.

I see that @IvanMalison had issue with #102 and said #93 should have fixed it, but I tried both the current version and I tried removing #102's change and neither worked when I had `package foo` and `package foo_test` in the same directory.


tehleach:
Opened #106 on a different branch

PR #111, opened 2016-11-11
Remove interface assertion

### LINKED ISSUE #110: import cycle
```
$ pwd
/Users/tejas/dev/src/github.com/NetSys/quilt/cluster/provider
$ mockery -name=EC2Client
Generating mock for: EC2Client
$ tree
.
├── amazon_test.go
├── awsSpot.go
├── cloud_config.go
├── constraint.go
├── gce.go
├── mocks
│   └── EC2Client.go
├── provider.go
├── provider_test.go
├── vagrant.go
└── vagrant_api.go

1 directory, 10 files

```

`mocks/EC2Client.go` is the generated file. That file is shown [here](https://gist.github.com/tejasmanohar/3c59bed42f3d46adba411c006e03bf42#file-go-L288). The last line `var _ provider.EC2Client = (*EC2Client)(nil)` causes an import cycle

```
$ go test ./cluster/provider
# github.com/NetSys/quilt/cluster/provider
import cycle not allowed in test
package github.com/NetSys/quilt/cluster/provider (test)
        imports github.com/NetSys/quilt/cluster/provider/mocks
        imports github.com/NetSys/quilt/cluster/provider

FAIL    github.com/NetSys/quilt/cluster/provider [setup failed]
```

I assume this line is to test that the interface is satisfied by the mock. That seems like a good idea, but do you have any suggestions on how to avoid to import cycle? Should the default be changed from generating in `./mocks`?

bogdanovich:
Fixes #110

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -249,10 +249,6 @@ func (g *Generator) GeneratePrologueNote(note string) {
 	}
 }
 
-func (g *Generator) GenerateInterfaceAssertion() {
-	g.printf("\nvar _ %s = (*%s)(nil)", g.renderType(g.iface.NamedType), g.mockName())
-}
-
 // ErrNotInterface is returned when the given type is not an interface
 // type.
 var ErrNotInterface = errors.New("expression not an interface")
--- a/mockery/walker.go
+++ b/mockery/walker.go
@@ -119,8 +119,6 @@ func (this *GeneratorVisitor) VisitWalk(iface *Interface) error {
 		return err
 	}
 
-	gen.GenerateInterfaceAssertion()
-
 	err = gen.Write(out)
 	if err != nil {
 		return err


IvanMalison:
I think there are better solutions to #110

bogdanovich:
@IvanMalison There are, definitely. This is a hotfix to unbreak the mockery until better solutions are implemented.

tejasmanohar:
Agreed with @bogdanovich. `master` has a regression right now.

codeactual:
@IvanMalison I've been trying to do more black box testing in my own projects, so personally I don't have a current need for this PR. But I do think mockery as a whole might have more value generally if white box testing as described in #110 was supported (again).

In #110 you mentioned:

> Yeah I added that to ensure that the mock satisfies the interface, so that if you do end up changing something, you will get very obvious output when you try to compile and you will know that you need to regenerate the mock.

I think the addition of that harness is quite valuable, but less so if it's non-negotiable. Non-negotiable, as breaking changes are, it added some value and then immediately removed some (for that white box testing case). 

If this behavior were controllable by a flag (I have no strong feelings about default-on vs. default-off), by making it negotiable and non-breaking, I think _it would only add value_.

Currently mockery: 

- does not support that white box testing scenario
- and does not support target interfaces that import vendored dependencies (unless `-dir` is an absolute path)

But it used to support both. Both have straightforward PRs that add back support. This is one of them.

(Also, @bogdanovich and @IvanMalison, I'm not clear on what those "better solutions" might be. Help?)

I'd absolutely like to see this feature kept long-term, but without it breaking an entire class of usage that's appropriate for some projects. Can we temporarily revert the feature until those "better solutions" can be explored?

PR #113, opened 2016-12-07
Move a note in the prologue to top of a file

kyoh86:
This request moves "notes in the prologue" (by #27) to top of each files.

This can be useful to collapse diffs of generated file in GitHub reviews.

For example: 

```
mockery -all -note "Code generated by mockery. DO NOT EDIT!"
```

Results that generated will look like this

```
// Code generated by mockery. DO NOT EDIT!

package mocks

import "github.com/stretchr/testify/mock"
```

and in GitHub reviews, like this:


kyoh86:
@evanphx 🙏

Issue #112, opened 2016-11-29
Loader changes break vendoring


tehleach:
Just FYI:

We noticed that generating mocks on files that rely on vendored dependencies broke yesterday. Looking through this packages non built-in dependencies, I found [some changes](https://github.com/golang/tools/commit/6c803ab014286e086881c410cba381b02317423c) made to [golang.org/x/tools/go/loader](http://golang.org/x/tools/go/loader) which seem to have shifted absolute path -> relative path. I spent some time trying to make mockery work with these changes, but just ended up vendoring the old commit on [my branch](https://github.com/tehleach/mockery) for now, couldn't figure out how to get it working.

mmcloughlin:
If you're using [glide](https://github.com/Masterminds/glide), this will get you a working version of mockery:

```yaml
- package: github.com/vektra/mockery
  version: 4db903efa8f6a978c8ccee3d4f23e947af878ca9
- package: golang.org/x/tools
  version: c060f04f93f57c44284360be78bb323077183645
  subpackages:
  - go/loader
```

nikolay-turpitko:
@mmcloughlin, Thank you Michael!

For workaround it Travis build I've replaced

```
go get github.com/vektra/mockery/.../
```

with

```
- glide up
- go install github.com/letsrock-today/authkit/vendor/github.com/vektra/mockery/.../
```

mmcloughlin:
@nikolay-turpitko I'd recommend using `glide install` in your build.

nikolay-turpitko:
@mmcloughlin yes, you are right, thanks for the tip

Franky12:
Is this issue fixed?

codeactual:
@Franky12 I don't think it is.
@tehleach Can this one be closed as a duplicate of #91?

tehleach:
@codeactual I don't think so - I'm pretty certain the issue began specifically with the commit to golang.org/x/tools/go/loader that I linked in the issue originally, which happened in early November. #91 is from June so I'm thinking it's unrelated. I guess I'm not really sure what the proper convention is for duplicates, but I feel like the solution to this issue will have something to do with absolute vs relative path so I think the info / discussion here would be useful.

codeactual:
@tehleach Thanks. I took a closer look and your hunch about absolute/relative was right on AFAICT. I hope the PR fixes it for you.

evanphx:
Hopefully #117 fixed this for you, please indicate if not.

Issue #91, opened 2016-06-23
The latest mockery does not recognize packages in vendor directory


any1fish:
When I'm trying to generate mocks, it seems latest mockery does not go into vendor folder for some packages:

```
Generating mocks...
wrappers/static.go:7:2: could not import github.com/shawnburke/godi (cannot find package "github.com/shawnburke/godi" in any of:
    /usr/local/Cellar/go/1.6.2/libexec/src/github.com/shawnburke/godi (from $GOROOT)
    /Users/chen/project/pyramid/src/github.com/shawnburke/godi (from $GOPATH))

```

The shawnburke/godi package indeed exists in vendor folder of my project, but it seems mockery does not search inside the vendor folder.

The following version of mockery does not have this issue and generates mocks successfully, though:

```
commit 84bc9c2423f94383b2d3b3571def7f8032962664
Author: Mike Gaffney <gaffo@justin.tv>
Date:   Wed Apr 6 13:53:06 2016 -0700

    Fix recursion break in #59
```

nathany:
The issue persists. Testing on commit 53e21da5f9d000ffafde06fec5e5ab169a5c42c6.

codeactual:
@nathany Hi, Nathan. What CLI flags and golang version were involved? I'm curious if they included `-dir` and a relative path. If so, #117 might cover your use case.

nathany:
@codeactual Nope, not using `-dir`:

```
mockery -name=Service -recursive
```

```
❯ go version
go version go1.8rc2 darwin/amd64
```

codeactual:
Thanks @nathany. I was able to reproduce your `-name` and `-recursive` case in `go1.8rc2 linux/amd64`. 

It looks like the default `-dir=.` value leads to the same problem as in #117: relative paths are passed to `Parser.Parse` which are incompatible with eventual dependence on `build.Context.Import`.

@evanphx Can PR #117 get a review? It should restore out-of-the-box vendor support that's been broken since golang 1.6.

codeactual:
@nathany Is that issue now resolved in master?

nathany:
It is certainly looking through vendor when using master (e4fa38885187a4dd421cbe250b50f3fc566a087a) [go version go1.8.1 darwin/amd64].

`mockery -name=Service -recursive` (without `-dir`) is taking a very long time, and producing many messages like:

```
[...]

PR #117, opened 2017-01-07
Fix support for relative paths to mock targets with vendor deps

codeactual:
Currently `-dir=./somelib` is not supported if `somelib` interfaces have vendor dependencies.

This PR may address the use cases in #91 and #112.

Explanation:

To support relative paths to mock targets w/ vendor deps, we need to provide eventual calls                                                                                                                                
to `build.Context.Import` with an absolute path. It needs to be absolute because
`Import` will only find the vendor directory if our target path for parsing is under
a "root" (`GOROOT` or a `GOPATH`). Only absolute paths will pass the prefix-based validation.

For example, after `Parse("./somelib")`, `Import` will check if any "roots" are a
prefix of `"somelib"` and decide to skip the vendor search.

It's possible that both the `go/loader` commits referenced/linked in #91 and #112 led to compatibility
issues with the [migration to go/loader](https://github.com/vektra/mockery/commit/79510c38075c14feb801e9ce7e1eafe7792de80e). I think the current incompatibility, for the `-dir=./somelib` use case, is with `go/build` behavior that [landed in Go 1.6](https://github.com/golang/go/commit/0c428a56176353d52170f318e998f342b08dacd2) which made absolute paths required to trigger vendor search AFAICT.

### DIFF
--- a/mockery/parse.go
+++ b/mockery/parse.go
@@ -25,6 +25,20 @@ func NewParser() *Parser {
 }
 
 func (p *Parser) Parse(path string) error {
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return err
+	}
+
+	// To support relative paths to mock targets w/ vendor deps, we need to provide eventual
+	// calls to build.Context.Import with an absolute path. It needs to be absolute because
+	// Import will only find the vendor directory if our target path for parsing is under
+	// a "root" (GOROOT or a GOPATH). Only absolute paths will pass the prefix-based validation.
+	//
+	// For example, if our parse target is "./ifaces", Import will check if any "roots" are a
+	// prefix of "ifaces" and decide to skip the vendor search.
+	path = abs
+
 	dir := filepath.Dir(path)
 
 	files, err := ioutil.ReadDir(dir)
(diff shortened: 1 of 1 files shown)


mockery/parse.go
```diff
@@ -55,10 +69,6 @@ func (p *Parser) Parse(path string) error {
 		p.nameToASTFiles[f.Name.String()] = append(p.nameToASTFiles[f.Name.String()], f)
 	}
 
-	abs, err := filepath.Abs(path)
-	if err != nil {
-		return err
-	}
 	p.path = abs
```
nathany: Might be simpler to use `p.path = path` here.

mockery/parse.go
```diff
@@ -25,6 +25,20 @@ func NewParser() *Parser {
 }
 
 func (p *Parser) Parse(path string) error {
+	abs, err := filepath.Abs(path)
```
nathany: Rather than introduce another variable for people to think about, it'd be simpler to just update `path` here.

codeactual: Good call. Thanks.

PR #129, opened 2017-03-23
Add readme note about missing Return arguments

codeactual:
Based on the two recent issues related to that scenario.


jaytaylor:
I'd love to see this PR merged in - might've saved me some time =)

Issue #104, opened 2016-10-24
Files generated multiple times


atombender:
``` shell
$ mockery -name='CrudOperations|Mutation' -case=underscore
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
Generating mock for: CrudOperations
Generating mock for: Mutation
```

For some reason, Mockery is generating the mocks 9 times. There are 8 Go files in that package (excluding tests). If I add another one, Mockery generates mocks 10 times, so it's obviously tied to the number of files.

andymc904:
Did not seem to be behavior at commit `47a058062601e235731714f89d7ec95c1ec47a5e`

PR #133, opened 2017-03-27
Fixed performance issue caused by loading config multiple times

andymc904:
Fixed  performance issues caused by loading config after every file load

See attached cpu profiles for more info on performance issues. 
Measured cpu profile on `make integration` in both cases
[cpu-profile.zip](https://github.com/vektra/mockery/files/871272/cpu-profile.zip)

Also fixes https://github.com/vektra/mockery/issues/104


codeactual:
I think there are several opportunities to improve this for review-ability and scope.

- Remove as many file changes as possible which are not critical to the PR. GitHub says that 1,889 files were changed and 823,428 lines were added. A quick look at the manifest shows things like vendor tests, JS, and other non-Go code files.
- Consider splitting the `-cpuprofile` feature into a separate PR from the performance fix. I see how they were related in origin but they're easier to review apart.
- Package management seems endlessly controversial. I've used glide, worked well for me. But adoption of it might also fit better in its own PR that is only about packaging.

andymc904:
Fair enough I'll just get the relevant changes in here.

But FWIW vendoring is something that should seriously be considered

andymc904:
Added `cpuprofile` cli option PR https://github.com/vektra/mockery/pull/134

andymc904:
Added issue to discuss preferred vendoring tools https://github.com/vektra/mockery/issues/135

PR #138, opened 2017-04-01
Vendoring external deps using glide

tamccall:


codeactual:
If glide doesn't support vendor pruning [[1](https://github.com/golang/dep/issues/120), [2](https://github.com/golang/dep/pull/322)], it still might be worthwhile to do it via `find` in the `Makefile` ([example](https://github.com/golang/dep/pull/322#issue-213730765)) after each fetch.

This commit, and any future one to upgrade deps, will contain all sorts of files that aren't licenses or built (e.g. txt, javascript, etc.).

andymc904:
@codeactual added a glide up target in the `Makefile` based off of the example you gave that will perform `glide up` and remove unnecessary files

tamccall:
Ran into [this guy](https://github.com/sgotti/glide-vc) the other day would likely resolve @codeactual issues with unnecessary code in the vendor directory

evanphx:
This is an oldie. I probably should vendor them but I'd do so with dep now instead.

Issue #121, opened 2017-02-20
proposal: add semvar


alanchchen:
I suggest adding semvar to both codegen and the generated code.

E.g.,
```
$ mockery -version
1.0.0
```
```
// Code generated by mockery v1.0.0
// DO NOT EDIT!
```

andymc904:
@evanphx instead of just adding an additional [cli arg](https://github.com/vektra/mockery/blob/master/cmd/mockery/mockery.go#L106-L127), can we migrate this package to use [urfave/cli](https://github.com/urfave/cli)? It has support for the version as requested by this issue plus we'd get a `-help` command for free

kingofsouth:
@alanchchen A change to allow for this was merged in yesterday try merging in master to see if this issue can be closed.

PR #140, opened 2017-04-04
Versioned files

kingofsouth:
Addressing issue from https://github.com/vektra/mockery/issues/121


kingofsouth:
Should add test to make sure this comment ends up on the 0th line. To address https://github.com/vektra/mockery/issues/150

mockery/generator_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strings"
 
 	"github.com/stretchr/testify/suite"
+	"github.com/stretchr/testify/require"
+	"bufio"
```
kingofsouth: probably should clean up these imports

Issue #141, opened 2017-04-05
Regression in mock generation


curtisallen:
It looks like https://github.com/vektra/mockery/pull/133/commits/81f429d47bc8fa3d20e4136c270b6d3a12a44324 introduced a regression. Causing import cycles in generated mocks.

Given 

`foo/foo.go`
```go
package foo

type Data struct {
}

type Foo interface {
	Something() Data
}
```
The generated mock produces an import cycle using e4fa388

```console
mockery -name=Foo -dir=foo -inpkg
```
Generates a mock with
`foo/mock_Foo.go`
```go
import foo "github.com/curtisallen/mockery-bug/foo"
...
// Something provides a mock function with given fields:
func (_m *MockFoo) Something() foo.Data {
	ret := _m.Called()

	var r0 foo.Data
	if rf, ok := ret.Get(0).(func() foo.Data); ok {
		r0 = rf()
	} else {
		r0 = ret.Get(0).(foo.Data)
	}

	return r0
}
```

However using `370e224` the mock generated doesn't have an import cycle. 

I've created a repo demonstrating the bug  https://github.com/curtisallen/mockery-bug
See [correct mock](https://github.com/curtisallen/mockery-bug/blob/master/foo/GOOD370e224_mock_Foo.go)
Import cycle [mock](https://github.com/curtisallen/mockery-bug/blob/master/foo/mock_Foo.go)

cc:/ @kingofsouth

kingofsouth:
I found this same issue at work a few days ago.  I have a change locally to fix it [here](https://github.com/kingofsouth/mockery/tree/inpkg-bug), but I want to write some tests to ensure it doesn't happen again.

curtisallen:
Great thanks!

kingofsouth:
I went ahead and put up a pr for my fix in an attempt to get master back into a good state,

Haven't been able to repro in unit testing yet

kingofsouth:
@evanphx Could https://github.com/vektra/mockery/issues/143 get a review this issue has been open for quite awhile, and the tip of master is currently broken.

wolfgangmeyers:
[smashingpumpkins.zip](https://github.com/vektra/mockery/files/1005141/smashingpumpkins.zip)
I've attached an easy reproduction scenario for this issue, which I believe is the same one I've run into. If not let me know and I'll open up a new one :)

kingofsouth:
[...]

PR #143, opened 2017-04-08
Fixed bug in inpkg where generated source would import itself

kingofsouth:
Fixes import cycle bug when doing inpkg generation

See: https://github.com/vektra/mockery/issues/141

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -240,9 +240,13 @@ func (g *Generator) sortedImportNames() (importNames []string) {
 }
 
 func (g *Generator) generateImports() {
+	pkgPath := g.nameToPackagePath[g.iface.Pkg.Name()]
 	// Sort by import name so that we get a deterministic order
 	for _, name := range g.sortedImportNames() {
 		path := g.nameToPackagePath[name]
+		if g.ip && path == pkgPath {
+			continue
+		}
 		g.printf("import %s \"%s\"\n", name, path)
 	}
 }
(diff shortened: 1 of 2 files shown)


kingofsouth:
@codeactual @evanphx could i get you guys to take a look at this? Trying to get master into a better state

codeactual:
Future commits for #144 probably belong in this PR instead because the new tests would verify the fix.

Does inpkg need more coverage, unrelated to import cycles, to assert its general behavior after this patch?

One initial thought on the commit: some comments on the changed `if` lines that were changed might be helpful, e.g. "X always needs to be compared to Y to prevent import cycles", etc. in case those lines need to be revisited later for some other purpose.

kingofsouth:
`Does inpkg need more coverage, unrelated to import cycles, to assert its general behavior after this patch`
I was mainly refering to this issue

`X always needs to be compared to Y..`
 I'll add some comments

`Future commits for #144 probably belong in this PR instead because the new tests would verify the fix.` 
are you suggesting that we hold off on the fix until we add enough testing to verify it?

codeactual:
> I was mainly refering to this issue

I was curious if, in your opinion, more tests are needed in general to reduce inpkg regressions (that might even result from this fix). I've only taken a closer look a very small portion of the GeneratorSuite, so I don't have an opinion. This issue raised the question of whether #141 was sort of bound to happen at some point because there wasn't good enough coverage historically. 

> are you suggesting that we hold off on the fix until we add enough testing to verify it?

I think there's plenty of time for it.

 You were [already quick to provide a patch](https://github.com/vektra/mockery/issues/141#issuecomment-292018393) for anyone with an urgent need. I think it's reasonable for everyone else to wait for a verified one.

kingofsouth:
added a test to catch the import cycle

kingofsouth:
> I was curious if, in your opinion, more tests are needed in general to reduce inpkg regressions (that might even result from this fix). I've only taken a closer look a very small portion of the GeneratorSuite, so I don't have an opinion. This issue raised the question of whether #141 was sort of bound to happen at some point because there wasn't good enough coverage historically.

I do agree with the statement that "#141 was sort of bound to happen at some point," so adding additional testing for inpkg generation is likely needed. However I am not sure how we would want to about adding those tests. There are about a million different cases that are being tested by that suite, so I certainly wouldn't advocate for verifying every case again for inpkg mode.

What might make more sense to me would be adding an inpkg run to the integration target and asserting that the package still builds after running mockery

codeactual:
Yep. I was not advocating to multiply the coverage or something ("million different cases") for the inpkg scenario.

This was my only concern:

> I was curious if, in your opinion, more tests are needed in general to reduce inpkg regressions (that might even result from this fix). 

If the new test is sufficient, that is ideal.

evanphx:
Sorry about the delay! Thanks for the fix

Issue #119, opened 2017-02-14
Cannot generate mock for interface: string literal not terminated


yuderekyu:
I've been trying to generate mock interfaces, but have been running into issues.

My interface (SubscriptionI) builds just fine, but fails when trying to generate a mock. 

```golang
type SubscriptionI interface {
	GetByID(string) (*models.Subscription, error)
	GetAll(int, int) ([]*models.Subscription, error)
	GetByRoaster(string, int, int) ([]*models.Subscription, error)
	GetByUser(string, int, int) ([]*models.Subscription, error)
	Insert(*models.Subscription) error
	Update(string, *models.Subscription) error
	SetStatus(string, models.SubscriptionStatus) error
	Delete(string) error
}
```
I get the error:
>Generating mock for: SubscriptionI
>Error walking SubscriptionI: mock.go:3:16: string literal not terminated (and 9 more errors)

Is it possible there is some parsing error?

codeactual:
Hi @yuderekyu,

I feel like I've seen that error before but I can't recall what triggered it. Can you provide the CLI flags that were used, golang version, and the partially generated mock (if one was output)?

yuderekyu:
CLI flags: `mockery -dir=helpers -name=SubscriptionI -output=./_mocks/handlers`
Version: `go1.7.3 windows/amd64`

Mockery generates a blank `SubscriptionI` in the desired output folder, `_mocks/handlers`.

codeactual:
Thanks for the info.

> Is it possible there is some parsing error?

I think that's correct and may be related to what's reported in #14. `imports` is used to fix the imports just before the mock is written to file, on [generator.go L552](https://github.com/vektra/mockery/blob/53e21da5f9d000ffafde06fec5e5ab169a5c42c6/mockery/generator.go#L552), but `imports` encounters bad syntax that `mockery` generated.

I wasn't able to reproduce the problem on `go1.8rc3 linux/amd64` with the provided flags/code and a `models` package with empty structs. My guess is that it's a Windows-specific issue.

It would help to know what [g.buf.String() in Write()](https://github.com/vektra/mockery/blob/53e21da5f9d000ffafde06fec5e5ab169a5c42c6/mockery/generator.go#L550) looks like on Windows in order to know how to fix the syntax.

codeactual:
@yuderekyu In case you have time to test a patch, #124 may fix address your parsing error.

codeactual:
@yuderekyu Thanks.

> Updated mockery.

[...]

PR #149, opened 2017-04-13
Print mock.go to stderr on imports.Process error

codeactual:
Let people know what "mock.go" is when it appears in error messages, and provide the source so it can be debugged or pasted into bug reports.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -631,6 +631,8 @@ func (g *Generator) Write(w io.Writer) error {
 
 	res, err := imports.Process("mock.go", theBytes, opt)
 	if err != nil {
+		line := "--------------------------------------------------------------------------------------------"
+		fmt.Fprintf(os.Stderr, "Between the lines is the file (mock.go) mockery generated in-memory but detected as invalid:\n%s\n%s\n%s\n", line, g.buf.String(), line)
 		return err
 	}
 


mockery/generator.go
```diff
@@ -631,6 +631,8 @@ func (g *Generator) Write(w io.Writer) error {
 
 	res, err := imports.Process("mock.go", theBytes, opt)
 	if err != nil {
+		line := "--------------------------------------------------------------------------------------------"
+		fmt.Fprintf(os.Stderr, "Between the lines is the file (mock.go) mockery generated in-memory but detected as invalid:\n%s\n%s\n%s\n", line, g.buf.String(), line)
```
kingofsouth: I feel like this comment might be a little self evident and maybe even redundant assuming that the error message gets printed elsewhere.

It should be easy enough to figure out that the when I see an error message, and right above it is a whole go file printed to my console that that is the go file that is causing issues.

Just my opinion,

codeactual: I didn't know what `mock.go` was until looking in the source. The errors that I've encountered at that site have all mentioned `mock.go`. I couldn't see what's inside `mock.go` to associate line numbers with the source that caused the error.

The whole motivation for this was ambiguity/opaqueness. I'm comfortable if this change swings to the other end and appears to over-communicate.

kingofsouth: I'm mainly referring to the `Between the lines is the file (mock.go) mockery generated in-memory but detected as invalid:` comment I think printing the source is a great idea and something I've hacked in before to figure out what was going on

codeactual: Yep. I was addressing your main concern. I also mentioned what "additional information" I wanted to convey:

> I didn't know what mock.go was until looking in the source. The errors that I've encountered at that site have all mentioned mock.go.

The message explains [what mock.go is](https://github.com/vektra/mockery/issues/119#issue-207597729). We agree it doesn't help with debugging (future work). Instead, it just clarifies what just went wrong (past event).

Put another why: it explains why mockery just stopped and didn't do the work that was expected.

Is that TMI?

kingofsouth: I think it would be pretty apparent with or without that line, but that is just my opinion

codeactual: Thanks man.

PR #153, opened 2017-05-16
Add "-quiet" flag that suppresses output to stdout when set

### LINKED ISSUE #152: Add "-quiet" flag
I'm running `mockery` several times as part of a `go generate` task, and the `Generating mock for: ` output is not helpful in this case. It would be helpful to have a `-quiet` flag that would suppress output to `stdout` such that a successful run would not print anything.

nmiyake:
Fixes #152

### DIFF
--- a/cmd/mockery/mockery.go
+++ b/cmd/mockery/mockery.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/vektra/mockery/mockery"
 	"runtime/pprof"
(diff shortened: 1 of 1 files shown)


evanphx:
You included the build in here and it needs conflicts. Can you remove the binary and fix the conflict?

nmiyake:
Sorry about that -- fixed.

PR #156, opened 2017-05-23
Make prologue note lint-compliant

### LINKED ISSUE #155: Prologue note creates lint artefacts
With the addition of the prologue note, the generated code no longer passes standard for godoc:

```
   Error  (GOLINT) GOLINT
    Package comment should be of the form "Package mocks ..."

    >>>        1 // Code generated by mockery v1.0.0
               2 package mocks
               3 
               4 import mock "github.com/stretchr/testify/mock"
```

I would suggest changing to:

    Package mocks contains mocks generated by mockery [version]

Willyham:
Resolves #155


kingofsouth:
You are going to need to modify this before your build will pass.
https://github.com/vektra/mockery/blob/master/mockery/generator_test.go#L267-L278

evanphx:
This comment isn't true though 😄 . Users can opt to write the mocks into the current package.

kingofsouth:
Wherever we end up on this string I would suggest adding some test to verify the behavior here https://github.com/vektra/mockery/blob/master/mockery/generator_test.go

Willyham:
@evanphx Ah, you're right. This is slightly more complex than I thought... will re-evaluate when I get some time soon.

kingofsouth:
@Willyham @nmiyake  had some comments on your issue around the contents of that string

nmiyake:
Specifically, I think the only functional change required is to move line 272 (`g.printf("\n")`) before the `if note != "" {` block to ensure that there's always a newline printed after the comment (whether or not it's followed by a note). This will ensure there's a newline between the comment and the `package` clause, which will fix the linting issue with minimal changes.

ejj:
I'm running into this myself as well.  Anything I could do to facilitate it getting merged?  I'd be happy to fix up the PR if that would help.  Please let me know.

tamccall:
@ejj looks like there are still some comments on the best way to do this see: https://github.com/vektra/mockery/issues/155#issuecomment-304962473

Also, that unit test such that is comparing the generated file to a [hard-coded string](https://github.com/vektra/mockery/blob/master/mockery/generator_test.go#L257-L262) so that hard coded string will need to be changed before this will pass.

Just wanted to put that out there in case you decide to try and fix it

ejj:
Sure, I'll give it a go.

evanphx:
This is fixed by #169

PR #158, opened 2017-05-25
Fix quiet flag

nmiyake:
Quiet flag implementation previously had issue where output would
bleed into generated file. Setting stdout to nil resolves the issue.


nmiyake:
Tested locally to verify that this fixes the issue.

Writing a test for this would require doing a `go run` or compile-and-run of source as a separate process (since it has to do with redirecting the output of `stdout`). I can certainly write such a test, but it would involve a bit of overhead (create new temporary directory, create interface, do a `go run` on `main.go`, then verify output and clean up), so would want to make sure that such a test is desired/would be accepted before writing it.

nmiyake:
@evanphx

LandonTClipp:
Need to edit tests to get CI to pass.

LandonTClipp:
This is going to be fixed by just simply letting all log output go to stderr, and anything else to stdout (if needed).

PR #167, opened 2017-06-27
Tell golint to ignore mockery output

ejj:
The golint commit a5f4a24736 (Ignore generated code) causes it to
ignore files that begin with "Code generated" and end with "DO NOT
EDIT.".   By adding those lines to the prologue, we cause suppress
any warnings golint may flag on the generated code.  This patch also
adds an additional newline between the prologue and the Package
statement, which makes it clear to any other elements of the go
tool-chain that the prologue is not the package comment.

Fixes #155

PR #169, opened 2017-08-09
Lint friendly comment

uudashr:
Always provide empty line for non package documentation.
This can avoid the lint check since the document is not a package documentation.


uudashr:
This is supposed to solve #155 
Override the PR #167 and also #156

uudashr:
@evanphx  believe we can merge some PR

PR #175, opened 2017-11-07
Adding new KeepTree flag - allows to preserve tree struct for mocks

corentone:
* Using the keeptree flag allows to generate mocks in a different
directory but keeping the original structure of the repo that is being mocked.

* Very useful for kubernetes client-go that has sometimes 3 versions of the
same interface name hence requires different folders. Easiest way is to keep the
same tree structure as if it was "inPlace".

* KeepTree flag is most useful when used with -all, -dir and -output.


corentone:
I'm still testing it a bit more but would love to hear if this is something that could benefit more people.
It could use a bit more testing but I couldn't find much tests for the outputter logic.

corentone:
@evanphx any opinion?

JyBP:
I have several interface with the same name clashing .

I tried using your branch with the keeptree flag but nothing changed. I created a simple repository to reproduce the issue:
https://github.com/JyBP/keeptree

https://github.com/JyBP/keeptree/commit/61989286d71e2f55b13d75e2baecb8a1ed46c60b

Is your PR supposed to fix those kind of conflicts as well?

corentone:
@JyBP sorry for the delayed reply.

This worked for me:
https://gist.github.com/corentone/99a0d6f540a9afbe2c90f520ed08a8e6

KeepTree is good when you want to bulk generate all the mocks.
In your case, if you want to keep two separate commands, you may be able to use, from master:
```
mockery -name Dep -dir ./dep1  -output ./cmd/mocks/dep1
mockery -name Dep -dir ./dep2 -output ./cmd/mocks/dep2
```

PR #179, opened 2018-01-03
Modify disclaimer to align with Go standard

VeryCB:
There is a new standard about generated file disclaimer. According to https://golang.org/s/generatedcode

```
Generated files are marked by a line of text that matches the regular expression, in Go syntax:

^// Code generated .* DO NOT EDIT\.$

The .* means the tool can put whatever folderol it wants in there, but the comment must be a single line and must start with Code generated and end with DO NOT EDIT., with a period.

```

Generated files with a disclaimer match above regular expression will be ignored by golint as well.


VeryCB:
duplicated with #169

PR #190, opened 2018-03-19
Add "New" method for generated mock

posener:
The new method accpets a *testing.T object, It initiate the
mock structure and assign to it the test object with the new Test()
function that mock.Mock now supports.


evanphx:
Given it just creates the empty struct to pass to Test, I think I'd rather not add that. It's too easy for folks to then need to pass things to `New()` to construct the object, etc.

posener:
I am sorry to see you dismissing this PR.
Let me explain a bit more why you should accept it.

It initiate the mock object with the test field - so one could write:
```
m := NewMyMock(t)
```
instead of
```
var m MyMock
m.Test(t)
```
Which is much nicer, specially when you have several mock objects that you initiate in one test.
I don't see a reason not to have it.

When adding the test field to the mock, instead of panicking, it fails the test, which produces much more readable test results.

I hope I convinced you to change your mind.
Please provide informative information why not accepting this PR.

PR #195, opened 2018-04-09
Allow to pass to Mock.Return one func that return all out params.

### LINKED ISSUE #194: Allow to "Return" one func returning all out params
Now it's required to pass one function per out param, but usually it's much simpler to return all out params at once

skipor:
Allow to pass to Mock.Return one func that return all out params.
Fixes #194


skipor:
how to regenerate fixtures?

evanphx:
This still doesn't pass CI.

LandonTClipp:
I'm not sure I understand what the need is for this. You want to be able to pass a function whose output will be passed to Mock.Return? What need does this serve?

Willyham:
I believe the idea would be:

```
foo := &mockFooer{}
foo.On("Bar", "something").Return(func(string) (Baz, error) {
   // ...
   return Baz{}, nil
})
```

I'm not exactly sure what the use cases would be though. I think it would be nice to see some situations where this simplifies things.

skipor:
Main reason - simple mock setup to return different results on same method. Now they may depend on passed arguments and some state out of mock.
for example, if you need check that client retry requests, mock may be configured like this
```go
foo := &mockFooer{}
var n int
foo.On("Bar", "something").Return(func(str string) (*Baz, error) {
   n++
   if n < 3 {
      return nil, errors.New("temporary fail")
   }
   return &Baz{}, nil
})
```

Extra reason - it is really useful, when you need some complex assertions on arguments passed to mocks.
Yes, it's also possible to do it using `Run` but that requires many type assertions.

LandonTClipp:
I just realized this is already a feature, is it not?

https://github.com/vektra/mockery/commit/59ab6b5153d59cb24e5b9472a6bec1c093c3d382

LandonTClipp:
@skipor I'm not entirely convinced of the usefulness of this feature. It seems to me like almost all cases where you use this type of logic could be solved with how you chain the expectations. For instance your example is equivalent to this:

```go
foo.On("Bar", "something").Return(&Baz{}, nil).Times(2)
foo.On("Bar", "something").Return(nil, errors.New("temporary fail")).Once()
```

It's also a bit of a misnomer because intuitively your example should return that actual function, not call it and return the args. I think if you want something like that, what #247 is trying to do is a bit better.

I'm not opposed to it in principle, I just need to be shown an example where it can't be done with the already existing logic and would greatly simplify things.

LandonTClipp:
And also I'm a bit confused, isn't this behavior already implemented?

https://github.com/vektra/mockery#return-value-provider-functions

PR #200, opened 2018-05-29
Support conventional "snake" as a case argument value

andrewslotin:
I've been trying to use `mockery` for one of my projects and as a first time user found it counter-intuitive that it uses `underscore` as a case name instead of a conventional `snake`.

My bad, I didn't read the `README` first and was trying to use it guided by the help message provided by `mockery --help` 😃Though I believe it supporting both case names would make its use more intuitive and user-friendly.


evanphx:
Thanks!

PR #208, opened 2018-08-07
Use default GOPATH if env not set

uudashr:
This should remove necessity to set GOPATH env, and will use the default one if env not set.


uudashr:
@evanphx or @codeactua need your help on this PR. Please review

PR #216, opened 2018-10-15
include compile-time type checks in output

thepwagner:
I want to be sure my generated mocks are in sync with the base
interface.

Following the pattern from https://golang.org/doc/effective_go.html#blank_implements


codeactual:
Some people have hit import cycles (https://github.com/vektra/mockery/issues/110) when the assertion is always there. 

It was suggested in that thread to add a flag. If it's off by default, then it shouldn't suddenly break anything for anyone.

thepwagner:
@codeactual thanks for that link!

I don't think import cycles are possible in this implementation. Since it guards on the `-inpkg` flag, the generated file's imports are the same as the interface file's(?)

I appreciate defensiveness though: I'll refactor to a unique flag, and document that usage without `-inpkg` may generate import cycles.

thepwagner:
@codeactual I've refactored so assertions are only added when the `-typeassert` flag is added.

One thing we've noticed:
The file filter https://github.com/vektra/mockery/blob/ea265755d541b124de6bc248f7744eab9005fd33/mockery/parse.go#L72 is agnostic to the output file format https://github.com/vektra/mockery/blob/ea265755d541b124de6bc248f7744eab9005fd33/mockery/outputter.go#L71.
This means mockery will scan for the code that mockery has generated in previous runs, making type assertions volatile (e.g.  after adding a method to `Foo` interface, mockery will scan+explode on `mock_foo.go` since the type assertion now fails).

I think that warrants a separate issue, `-typeassert` just exacerbates it.

codeactual:
@thepwagner Good catch. A separate sounds good.

PR #217, opened 2018-11-01
Add flags to specify generated struct and file names

pohang:
When a single mock is generated (non regex provided to -name),
can optionally specify -filename and/or -structname to override
the default names.


kohago:
@pohang  Great! I just want to use that feature ASAP!

freedemster:
This is a very useful feature, I would like to use it too.

pdettori:
this is indeed a very useful feature ... any chance to get this PR moving forward ?

LandonTClipp:
I think this is a great feature. Let's hold off on this until I can get CI auto-tagging. After that I will merge it.

PR #218, opened 2018-11-13
Support Go 1.11 modules

### LINKED ISSUE #213: Fails with import errors with modules on Go 1.11
If I have a simple file such as this that refers to an imported package:

```
package mypackage
import "github.com/jackc/pgx"
// ...
```

then it fails with:

```
/Users/alex/.go/pkg/mod/github.com/jackc/pgx@v3.1.0+incompatible/batch.go:6:2: could not import github.com/jackc/pgx/pgproto3 (cannot find package "github.com/jackc/pgx/pgproto3" in any of:
	/usr/local/opt/go/libexec/src/github.com/jackc/pgx/pgproto3 (from $GOROOT)
	/Users/alex/.go/src/github.com/jackc/pgx/pgproto3 (from $GOPATH))
```

In other words, it's looking for the source in the old location, not in `$GOPATH/pkg/mod`.

I'm able to reproduce this in a small Go program that uses `golang.org/x/tools/go/loader` to load files, and I can confirm that it does not work with Go modules. (E.g. see [this issue](https://github.com/google/go-cloud/issues/78) for discussion.) Apparently the `golang.org/x/tools/go/loader` library has been superceded by [golang.org/x/tools/go/packages](https://godoc.org/golang.org/x/tools/go/packages), which is module-aware.

atombender:
This uses the new package `golang.org/x/tools/go/packages` instead of `golang.org/x/tools/go/loader` to load code, so that we can support checking packages that use Go modules with Go >= 1.11. This fixes #213.

This preserves a bunch of code that probably isn't needed anymore; `go/packages` supports the Go package pattern that `go list` uses (in fact, it appears to use `go list` internally), so all the file system walking could be removed in favour of just relying on that. The current code is inefficient, since doing `packages.Load(&conf, "file=./mypkg/myfile.go")` actually loads the entire package, and there's no need to then do `packages.Load(&conf, "file=./mypkg/otherfile.go")`, since it's already loaded.

There is one change here that I wasn't 100% sure about. The generator test cases have a test case for relative paths, something which I believe is moot with the new Go module system (and I'm not even sure if relative imports are officially supported in Go?). I have removed the test case in favour of using Go module imports instead of relative ones. Comments are welcome.

I added a `go.mod` file for good measure.

Note that this changes Mockery to require Go >= 1.9, as `go/packages` does not support earlier versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ language: go
 go_import_path: github.com/vektra/mockery
 
 go:
-  - 1.7.x
-  - 1.8.x
   - 1.9.x
   - "1.10.x"
   - tip
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,9 @@
+module github.com/vektra/mockery
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/stretchr/objx v0.1.1 // indirect
+	github.com/stretchr/testify v1.2.2
+	golang.org/x/tools v0.0.0-20181112210238-4b1f3b6b1646
+)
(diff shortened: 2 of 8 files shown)


evanphx:
Thanks!

Issue #220, opened 2018-11-19
Unable to get writer when used with -inpkg option


wulifu2hao:
Hi I encounter problem when using mockery with -inpkg option
how I use it: mockery -name IBookingService -case=underscore -inpkg
error msg: Unable to get writer for IBookingService: open gitlab.myteksi.net/gophers/go/food/external/mock_i_booking_service.go: no such file or directory

I suspect this is due to the recent change ( https://github.com/vektra/mockery/commit/2be70d1c16b905d1ebbb3f1da219b11de416a0e6#diff-9fccd60b1234c112849b2bdd764adae3R159 ): pkg.Path used to be the path of the file containing the interface, but now entry.pkg.Types points to the director of the file of the interface. I might be wrong

euskadi31:
ping @atombender

atombender:
I will take a look, thanks.

cicely-zhang:
Is there a way to get older version of mockery using go get to bypass this issue for now?

atombender:
You should be able to just clone the repo, `git reset --hard ea265755d541b124de6bc248f7744eab9005fd33` (for example), then run `go install .`.

atombender:
@wulifu2hao, @euskadi31, @santoshaitha, @cicely-zhang PR is available if you want to give it a go. Works for me, but there are no tests for the CLI itself, so I'm not as certain as I want to be that it's doing the right thing.

euskadi31:
it's working perfectly ;)

santoshaitha:
Is this PR merged?

PR #221, opened 2018-11-19
Fix regression with -inpkg

atombender:
Fixes regression (#220) in 0ecc4a91a394dcd6de2535c8262674f8b14767cb where `-inpkg` would get the wrong path.

Also cleans up some unused arguments and missing error handling in the same area of code. Renames `Interface` structs's `Path` to `QualifiedName` to avoid future confusion; with `go/packages`, this is not the file path, but the fully qualified package name.

### DIFF
--- a/mockery/generator_test.go
+++ b/mockery/generator_test.go
@@ -948,7 +948,7 @@ func (_m *Example) B(_a0 string) fixtureshttp.MyStruct {
 
 func (s *GeneratorSuite) TestGeneratorWithImportSameAsLocalPackageInpkgNoCycle() {
 	iface := s.getInterfaceFromFile("imports_same_as_package.go", "ImportsSameAsPackage")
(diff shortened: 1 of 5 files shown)


euskadi31:
Good for me 👍

it's working perfectly ;)

Thank You

julestruong:
@evanphx can you merge it ? it causes us lots of failure on our ci

santoshaitha:
Can this be merged ASAP? this causes lot of failures on our end. I think we should have tagging on this repository so that we can use latest stable tag when master branch broken

julestruong:
@santoshaitha +1 on tag or versioning

euskadi31:
@evanphx ?

evanphx:
Sorry for the delay folks!

euskadi31:
@evanphx Thank You !!

PR #234, opened 2019-05-10
Export mocks for unexported interfaces if not in package

### LINKED ISSUE #191: Add option to generate exported mock type for internal interfaces
#Hi,
Great tool, I really like it! Is there a possibility to add a new option like `-exported=true`, which will make internal interfaces exported?
Please consider an example:

```
//go:generate mockery -name=foo -output=automock -outpkg=automock -case=underscore -exported=true
type foo interface {
	Bar(name string) (string, bool, error)
}
```

It could generate a mock type in `automock` package with a name `Foo`, which is easy to use in test files.

**Alternative:**
It could generate a constructor for this type, which can be used from outside of the `automock` package.

stoewer:
When creating mocks for unexported interfaces the mocks become exported (starting with a capital letter) if `-inpkg` is not set.

This fixes #191 in a slightly different way as suggested in the original ticket. However generating unexported mocks in separate packages is rarely useful. Therefore I think this is still a good solution.


LandonTClipp:
If you set mockery to append `_test.go` to the mock filename, it doesn't matter that the mock becomes exported. Does that solve your issue?

PR #235, opened 2019-05-14
Utility function to extract function name

dudicp:
Providing a utility function to extract function names which is very useful when using mocks in order to make the unit-tests more robust for code changes i.e. rename of the function name


LandonTClipp:
I can merge this if you can get the CI to pass

README.md
```diff
@@ -190,3 +190,16 @@ Use `mockery -print` to have the resulting code printed out instead of written t
...
+
+### Extracting Function Names ###
+
+When using mocks, it's recommended to extract the function names using the below utility function in order to make your unit-tests more robust
+for code changes.
+
+```go
+import "github.com/stretchr/testify/utils"
+
+Mock.On(utils.GetFunctionName(mock.Function), mock.Anything).Return(func(s string) string {
```
LandonTClipp, line 202: Actually, I don't understand how this would make anything easier. If you end up changing the name of the function, you would have to modify your `Mock.On` call anyway, because the function name changed. Plus it looks like this `utils` package no longer exists.

dudicp: 1. After changing the function name, it will be catched during compilation which is better than during running the tests (and then figure out the function name have changed).
2. Changing the function name using IDE will automatically change this instance as well in contrary to string.

LandonTClipp:
That's a very good point! I think we can go ahead and merge this soon. I'll
set this as approved.

Regards,

--
Landon Clipp

README.md
LandonTClipp: That's a very good point. If you can fix the issue with testify/utils no longer existing (possibly find some suitable replacement), we can get this merged.

LandonTClipp:
Closing this PR due to inactivity.

PR #238, opened 2019-05-29
reduce the number of parsing

linzhp:
In [Walker.doWalk](https://github.com/vektra/mockery/blob/e78b021dcbb558a8e7ac1fc5bc757ad7c277bb81/mockery/walker.go#L80), it calls `Parser.parse` for each files in the directory. In [Parser.parse](https://github.com/vektra/mockery/blob/e78b021dcbb558a8e7ac1fc5bc757ad7c277bb81/mockery/parse.go#L59), it takes the directory of the given file, and calls `packages.Load` for all files in the directory again. If there are 10 files in the directory, mockery will eventually call `packages.Load` for 100 times!

This PR reduce the number of calls to `packages.Load` to once per file

### DIFF
--- a/mockery/parse.go
+++ b/mockery/parse.go
@@ -5,6 +5,7 @@ import (
 	"go/ast"
 	"go/types"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


LandonTClipp:
@linzhp can you rebase onto master to see if the tests pass?

linzhp:
Because of [this line](https://github.com/vektra/mockery/blob/255c6da5518cc05d918079453f02c372a53a5042/mockery/parse.go#L71), there is actually no duplication of `packages.Load` calls. However, it is still very expensive to call `packages.Load` (#258).

Closing this.

PR #242, opened 2019-08-23
Add [Deprecated] to the Readme

Edulopez:
Since there is no more support for the project I think it would be nice to add this mark.


LandonTClipp:
The project is being picked back up so this is not necessary anymore.

Issue #180, opened 2018-01-15
Publish docker image


pavolloffay:
Hi all, 

thanks for the awesome tool! Could you please publish docker images with this binary? It simplifies dev and CI usage.

evanphx:
I'm not sure why that's needed. If you're developing Go it's trivial to install.

rogchap:
Speed is why it's needed; you don't want to have to install mockery on every CI build, but rather have a cached version of the tool on the docker image.

LandonTClipp:
Closing historic ticket. Please open a new one if you still have issues.

rogchap:
It’s not been resolved with an appropriate answer @LandonTClipp; Would like to see a better reason to close, as an “offical” docker image is still required.

LandonTClipp:
Okay, looking back at this I think it'd be a good idea, however I don't have a lot of spare cycles to get this done at the moment. If you could submit code that would be great, then I could integrate with CI etc (which is still on my TODO list).

rogchap:
@LandonTClipp I can help put this together; including helping with CI. I recommend using [goreleaser](https://goreleaser.com/); which I can add as part of my PR.

LandonTClipp:
@rogchap goreleaser looks like a great option, could you add that in a separate PR to your docker PR? I will add any appropriate secrets into Travis.

rogchap:
@LandonTClipp please take a look at #257

LandonTClipp:
@rogchap your goreleaser PR is now working, we can tackle publishing docker images now. I assume this is an option supported by goreleaser?

rogchap:
Great, thanks @LandonTClipp 
Yes, adding docker will be straight forward now.
I also recommend adding a homebrew package. We can use a seperate repo for the tap. (Maybe as a seperate PR)

I noticed that the CI is trying to deploy 4 times because we are running in each Go env and after the first deployment the others fail. The easiest way to fix this is to only run with 1 Go environment (1.14).

LandonTClipp:
Yep I noticed that as well, you were right before, I think we should just
specify a single go version. Now there's a race between the builds to
upload the files to the release.

Regards,

--
Landon Clipp

rogchap:
We can still run the tests on all versions, but I’ll pin the deployment to just Go 1.14. PR incoming.

PR #257, opened 2020-05-08
Add configuration for goreleaser

rogchap:
This adds basic configuration for goreleaser. 
On a tagged release it will build for linux, osx and windows, and add to the GitHub release page, including the artifacts, changelog and checksum file.
To test locally you can run `goreleaser --snapshot --skip-publish --rm-dist`

I've also change the `SemVer` to be auto generated at build time via ldflags; This may cause slight confusion for people installing via `go get` as it will display `0.0.0-dev`, but allows easier version management via CI releases (with goreleaser); In the future we should favour CI builds over using `go get`.

This is a prerequisite to having other distributions such as docker (#180) and homebrew (recommended) 

I've amended CI (travis) to run the tests and do a deploy when there is a valid release tagged.


rogchap:
Looks like there is a issue with the golden files and line endings on Windows `\n` vs `\r\n` 😞 
Not sure the best way to fix? Maybe needs a build flag to select a different golden file based on platform? 

Rather than fix this now, I'm going to remove windows from the test script. We can still build for Windows though.

LandonTClipp:
Hi @rogchap , thanks a lot for the PR, this looks great. Unfortunately I think we really need to have the versioning saved into the go source code so that you can `go get` a specific version, as well as `mockery` being able to correctly identify its version when doing `go get`. The workflow I'd really like to have is this:
1. A human changes the `SemVer` variable in source, and commits to master.
2. Travis CI initiates build on master. Runs tests.
3. If tests are successful, Travis gets the SemVer using something like `mockery -version`
4. Tags the commit to master using the version from step 3, creates GH release.

I don't think that should be too difficult to implement. What do you think?

rogchap:
Hey @LandonTClipp,

I strongly recommend moving away from `go get` to install binaries. The Go Team have previously stated that this is a feature that they want to remove, but is only there for backward compatibility reasons.

`go get` results in inconsistent behavior depending on "gopath mode". If you use `go get` outside a Go modules repo this will install using "gopath mode" and would get, and install, the code in the `master` branch, which could have unreleased code. In "modules mode" this would install the latest git tagged version (unless the user does `go get ...@v1.0.0`). The version number installed is based off the git tag, not `SemVer`, so there is a possibility of version miss-matches.
Using a package manager for exec binaries are far easier to manage and maintain (homebrew etc); from both a user perspective and maintainer perspective.

goreleaser requires a git tag before it will build; so you always need a new valid git tag to do a deployment. The flow you describe is actually quite complex, because you would have to build the binary before building the binary (!!??) to get the version, check that there are no tags with that version and then (and only then) do a deployment based of that version. Not impossible, but quite messy, and unconventional from managing Go releases. 

You could `grep` (or something like that) to get the version; but this would be brittle. You also need to remember to always bump the version number in a Go file and then update the test (I guess we could change the test to read `SemVer`?). This will get tedious if every patch you merge you manually have to update master with a new version, rather than just doing a `git tag` and it being automated.

If you are not keen on master containing version `0.0.0-dev` maybe we could change this to `master` instead!?

If you really really really still want to maintain `go get` as a way to install, and really really really care that it does not output `master` (or `0.0.0-dev`) when doing `--version`; Then we can change the default `SemVer` to be `1.0.0` (or whatever is latest) but still have the binary build process "override" the value from the git tag (we should change the test so that the golden generated file matches correctly).
This would result in two manual steps: update `SemVer` then git tag with same version. If you ever forgot to bump the `SemVer` at least the built binaries would have the correct version embedded from the git tag.

.travis.yml
```diff
@@ -1,15 +1,27 @@
...
+os: 
+  - linux
+  - osx
+
+env:
+  - GO111MODULE=on CGO_ENABLED=0
 
 go:
-  - 1.11.x
-  - 1.12.x
```
LandonTClipp, line 8: stretchr/testify still supports 1.11/1.12. Yes I know they're ancient but let's keep them in here for now.

https://github.com/stretchr/testify#supported-go-versions

LandonTClipp:
@rogchap thank you for all your input. After thinking it over I think what you are saying makes sense and we can go ahead and move forward. I really don't think that it's that complicated of a workflow but if that's how most of the Go community releases binaries I think we should follow that.

I will look this over once more on Monday and get this merged. I'll put in a few minor comments. Thanks!!

.travis.yml
rogchap: Surly this only effects the build of the mockery binary, rather than the supported Go output by the tool? We would only build mockery binary with the latest (1.14) so we could actually _remove_ all but 1 Go version to run the tests. 💁‍♂️ 

That said, I've added both 1.11 and 1.12 back in as requested.

mnaboka:
oops, this change has broken our CI, we are installing latest mockery and check for version to be `1.0.0` :(

LandonTClipp:
Okay, we might just need to change the default version to be 1.0.0 for
backwards compatibility. I'll fix this.

Regards,

--
Landon Clipp

rogchap:
If you can run in modules mode, then `go get github.com/vektra/mockery/.../@1.0.0` should work?
Otherwise I would change your CI to not check for 1.0.0, but rather 1.x

LandonTClipp:
@mnaboka can you confirm your CI is fixed from the new commit?

mnaboka:
@LandonTClipp sure, i've fixed it on my end, another issue is that all generated mocks also include a string with version, so all those mocks will also be changed.
Thank you

LandonTClipp:
Yes the commit reverts back to the old behavior where SemVer is 1.0.0. Thank you for fixing it on your end. I've added a note in the README about this, we'll keep that deprecation for a while and then we can switch it back over to `0.0.0-dev` after a few months. After that people will just need to fix their CI.

PR #260, opened 2020-05-11
Add support for releasing an offical docker image

### LINKED ISSUE #180: Publish docker image

rogchap:
This adds a docker image plus the relevant docker goreleaser/travisci configuration.

To test locally run:
```zsh
goreleaser --snapshot --skip-publish --rm-dist
docker run vektra/mockery --version
```

To test more than just the version you will need to mount your volume:
```zsh
docker run -v "$PWD":/src -w /src vektra/mockery -all
```

**Note**: I originally used a `Scratch` base image, which puts the image around 6MB... but mockery makes use of the `go` command (`go list` etc), so I need to use an image with Go installed. This bumped the image to 375MB!! It's not a major problem as most people will have the Golang base imaged cached, but this could slow down peoples CI for downloading a larger image. 
2 future improvements would be to remove the need for executing `go` commands and installing `Go` onto the scratch image rather than using alpine.

@LandonTClipp We will need to add the secrets for travis: `DOCKER_USERNAME` and `DOCKER_PASSWORD`

I have presumed that there is/would be a docker repo for user `vektra`? We can change if needed.

I have not updated the README, as I think it would be best to get this CI pipeline confirmed working then update to suggest docker. I would recommend that for CI people use the docker version.

Closes #180

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -31,3 +31,20 @@ snapshot:
     name_template: "{{ .Tag }}-next"
 changelog:
     sort: asc
+dockers:
+  - goos: linux
+    goarch: amd64
+    binaries:
+      - mockery
+    image_templates:
+      - 'vektra/mockery:{{ .Tag }}'
+      - 'vektra/mockery:v{{ .Major }}'
+      - 'vektra/mockery:v{{ .Major }}.{{ .Minor }}'
+      - 'vektra/mockery:latest'
+    build_flag_templates:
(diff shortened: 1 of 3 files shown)


LandonTClipp:
I need to get added to the vektra dockerhub org before we can merge this, I've emailed @evanphx to get access to it.

PR #263, opened 2020-05-13
Add a homebrew deployment

rogchap:
This will deploy each release to a homebrew tap. This will allow developers to use `brew` to install and update the mockery binary. This works for both mac and linux. Eg:
```
brew install vektra/tap/mockery
brew upgrade mockery
```
@LandonTClipp This deployment will require that travisCI has access to the repo: `github.com/vektra/homebrew-tap`; this repo will also need creating.


LandonTClipp:
Hi @rogchap , thanks for the PR. I will have to ask @evanphx to give me access to create repos in the vektra organization. One moment.

PR #265, opened 2020-05-17
Adding pretty structured logging support

LandonTClipp:
Capture

This PR adds structured logging to mockery and replaces calls to fmt.Print* with zerolog. Structured part of logging is passed around via context.

I don't know if I like forcing functions to accept `ctx`, but that is how I generally pass structured logs around in my go projects. That's quite a long debate.


LandonTClipp:
I will leave this PR up for a week or two for comments.

LandonTClipp:
This PR is going to be part of the 2.0.0-alpha.1 tag

PR #266, opened 2020-05-18
Adding snapstore exporting to goreleaser

LandonTClipp:


LandonTClipp:
Having trouble getting the openssl comand to work. I think this has to do with the fact that I'm making the encrypted files in a fork instead of the actual repo (even when passing `-r` flag to travis command).

https://carlosbecker.com/posts/goreleaser-snap-travis/

PR #267, opened 2020-05-18
Run go mod tidy and clean up some unused fields/logic

xtonyjiang:
- Ran `go mod tidy`
- Removed some unused fields
- `Load()` no longer visits nodes concurrently with anything else (as of https://github.com/vektra/mockery/pull/218) so there's no point in using a waitgroup here anymore.


LandonTClipp:
This is probably going to conflict with a change I'll be making soon with logging. Let's hold off on this for just a bit.

Issue #210, opened 2018-08-13
"Mocking" a third-party interface


MikeWarren2014:
In the code under test, there are two methods of a `struct` that ultimately depends on a [`Session.Store`](https://github.com/astaxie/beego). When I tried to use mockery to generate a "mock" of that interface, with `mockery -name Store -inpkg %GOPATH%/src/github.com/astaxie/beego/session`, it tells me that it's `Unable to find Store in any go files under this path`. However, I can not only `cd` into that path, from where I was, but say `mockery -name Store -inpkg .`, and it generates the mock there. **I want it near my test code!**

snowzach:
I had the same issue... Use -dir instead of -inpkg
`mockery -dir %GOPATH%/src/github.com/astaxie/beego/session -name Store`

zserge:
How this is going to work with the new modules approach (when dependency sources are not stored in vendor or GOPATH)?

qJkee:
@zserge 
Just instead of `%GOPATH%/src`, use `$APP_PATH`
And final result will be(cd to app location)
`mockery -dir vendor/github.com/astaxie/beego/session -name Store`

zserge:
@qJkee Not sure I understand you. The concept of modules kind of deprecates `vendor` directories, so related modules are fetched somewhere in $GOPATH/pkg/mod and contain hashes in the directory names, thus making it hard to locate.

But it looks like by switching to `golang.org/x/tools/go/packages` resolver `mockery` now allows passing `-dir github.com/whatever/external/package/you/want` and it should just work.

qJkee:
@zserge 
Sorry, a little bit miss xD
Yea, new approach makes it harder, if you are not storing all dependencies in vendor dir.

nielskrijger:
@zserge sadly couldn't get that to work.
I tried a random example:

```bash
$ mockery -dir github.com/go-redis/redis -name Pipeliner
Unable to find Pipeliner in any go files under this path
```
(regardless of doing a `go get github.com/go-redis/redis`)

glyn:
Here's a workaround of sorts: https://github.com/projectriff/riff/blob/b07ffe0366ce7474a99db484c95ab245cbfb9cd1/Makefile#L31-L43

zombie-guru:
I usually create a file in my package that references the interfaces I need and generate using that.

```
package mocksiface

import (
)

type S3API interface {
  s3iface.S3API
}
```

tskillian:
[...]

PR #270, opened 2020-05-26
add source package support

vincent178:
Signed-off-by: Vincent Huang <vh7157@gmail.com>

## Background

There's a usecase we want to generate mock for external package, there's one solution in community https://github.com/vektra/mockery/issues/210#issuecomment-485026348, but it's not clean to specify another alias interface into my codebase. The other solution is using `-dir` like in this https://github.com/vektra/mockery/issues/210#issuecomment-521112586 but it requires some knowledge to use it right, like it's different under go module and `$GOPATH`.

With this change, we can generate an external package mock anywhere (with go.mod or under $GOPATH) with
```
$ mockery -srcpkg github.com/aws/aws-sdk-go/service/s3 -all
```

## Changes
* add flag `srcpkg`
* import `srcpkg`, read directory and set `BaseDir` to it

## Test
- [x] manually verified on local
- [x] add integration tests
- [x] add unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 mocks
 mockery.prof
 dist
+.idea
--- a/Makefile
+++ b/Makefile
@@ -18,13 +18,4 @@ install:
 	go install ./...
 
 integration:
-	rm -rf mocks
-	${GOPATH}/bin/mockery -all -recursive -cpuprofile="mockery.prof" -dir="mockery/fixtures"
-	if [ ! -d "mocks" ]; then \
-		echo "No Mock Dir Created"; \
-		exit 1; \
-	fi
-	if [ ! -f "mocks/AsyncProducer.go" ]; then \
-		echo "AsyncProducer.go not created"; \
-		echo 1; \
-	fi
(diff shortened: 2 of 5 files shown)


LandonTClipp:
Hi Vincent, this looks great! Can you add tests to verify that this works as expected?

vincent178:
Hi @LandonTClipp , PTAL

LandonTClipp:
@vincent178 according to the go/loader docs, it is deprecated: https://godoc.org/golang.org/x/tools/go/loader

```
Deprecated: This is an older API and does not have support for modules. Use golang.org/x/tools/go/packages instead.
```

Can we use `go/packages` instead? `go/packages` does not support modules.

vincent178:
@LandonTClipp good catch, it's interesting that I tested `go/loader` does support our usecase with go modules, but anyway, I have changed to `go/packages`, PTAL, thanks.

LandonTClipp:
Looks good to me. I was about to ask you to to add more unit tests but it looks like those don't really exist for main. This'll be something I'll fix in v2

PR #271, opened 2020-05-26
fix for argument type ...map[string]interface{}

kanaksinghal:
For the following function argument, the generator produces invalid code
```
Foo(arg1 ...map[string]interface{})
```

Whereas, it works fine for
```
Foo(arg1 ...Msi)
```

Reason: I think on Line #611 the type is checked to include `interface{}` which clearly is not enough.
So, the proposed fix is intended to be a safe workaround until this is looked over in depth.

### DIFF
--- a/mockery/generator.go
+++ b/mockery/generator.go
@@ -608,7 +608,7 @@ func (g *Generator) generateCalled(list *paramList, formattedParamNames string) 
 
 	var variadicArgsName string
 	variadicName := list.Names[namesLen-1]
-	variadicIface := strings.Contains(list.Types[namesLen-1], "interface{}")
+	variadicIface := strings.Contains(list.Types[namesLen-1], "interface{}") && !strings.Contains(list.Types[namesLen-1], "map")
 
 	if variadicIface {
 		// Variadic is already of the interface{} type, so we don't need special handling.


LandonTClipp:
Can you elaborate on the invalid code it's generating? What's the result?

kanaksinghal:
Ref: Line #617. It is needed for `map[string]interface{}` as well
```
		// Define _va to avoid "cannot use t (type T) as type []interface {} in append" error			
		// whenever the variadic type is non-interface{}.	
```

Currently, this happens 

```
// Foo provides a mock function with given fields: bar
func (_m *DBOrgManager) Foo(bar ...map[string]interface{}) error {
	var _ca []interface{}
	_ca = append(_ca, bar...) // <- cannot use bar (variable of type []map[string]interface{}) as []interface{} value in argument to append
	ret := _m.Called(_ca...)
	...
}

// Foo2 provides a mock function with given fields: bar
func (_m *DBOrgManager) Foo2(bar ...string) error {
	_va := make([]interface{}, len(bar))
	for _i := range columns {
		_va[_i] = columns[_i]
	}
	var _ca []interface{}
	_ca = append(_ca, _va...) // <- this works fine
	ret := _m.Called(_ca...)
	...
}
```

LandonTClipp:
@kanaksinghal if you write a unit test to ensure that this is indeed fixed we can get this merged. Thanks for the great find btw! I want to get this fixed.

PR #277, opened 2020-06-02
Towards #272

blaggacao:
This should reportedly fix it, wasn't able to properly build the docker image, though, to test. Sorry.

PR #278, opened 2020-06-02
A01 - Add docker e2e test

blaggacao:
Required to get #277 right / expected to fail

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
+FROM golang:1.14-alpine as builder
+
+COPY ./ /mockery
+RUN cd /mockery && go install ./...
+
 FROM golang:1.14-alpine
 
-COPY mockery /
+COPY --from=builder /go/bin/mockery /
 
 ENTRYPOINT ["/mockery"]
--- a/Makefile
+++ b/Makefile
@@ -17,5 +17,8 @@ fixture:
 install:
 	go install ./...
 
-integration:
+docker: install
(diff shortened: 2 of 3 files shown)


blaggacao:
@LandonTClipp Looks like something _is_ wring with the `Dockerfile`, the result consistently is the following:

```
$ make integration
...
standard_init_linux.go:211: exec user process caused "no such file or directory"
make: *** [Makefile:24: integration] Error 1
```

Could you please test on your machine, if it's something with my docker buildx?

blaggacao:
@LandonTClipp 

This is somewhat redundant to what
```
# .goreleaser.yml
dockers:
  - goos: linux
    goarch: amd64
    binaries:
      - mockery
```
tries to accomplish with 
```
builds:
  - main: ./cmd/mockery/mockery.go
    ldflags:
      - -s -w -X github.com/vektra/mockery/mockery.SemVer={{.Version}}
    env:
      - CGO_ENABLED=0
    goos:
      - darwin
      - linux
      - windows
    goarch:
      - amd64
```

But it's just so much safer...

Maybe the design here for goreleaser is not the best.
In my case I can confirm that the binary built on the host machine was not working properly within the container, that's why I named the commit "I'd blame glibc ... "


Could you therefore please merge this branch and keep on the lookout on how goreleaser would solve this issue eventually? -> so we can move on to #280

blaggacao:
Can I push for merging this? It blocks #280

PR #283, opened 2020-06-14
Enable adding prefix for the generated package name

larryzhao:
When I use mockery to walk through a large amount of code files with in a deep structured directories, if there are interfaces with same name located in different packages, it's not so nice to generate all under the same package `mocks`, and it would be hard to read/easy to make mistakes if we use the original package name, because it would be the same package name from different paths.

I think it would be better to allow to generat mock files with a prefix like `mocks_` followed by the original package name. It's more clear to locate interface with same name under different package, and also easier to refer to in code.


LandonTClipp:
This can probably just be solved by using `--inpackage` couldn't it? Or instead you can use `--keeptree` and if you need to import two interfaces with the same name but in different packages, just import the mock packages with a different alias name.

larryzhao:
> This can probably just be solved by using `--inpackage` couldn't it? Or instead you can use `--keeptree` and if you need to import two interfaces with the same name but in different packages, just import the mock packages with a different alias name.

`--inpkg` goes into the original package, but i want them all in a different package.

`--keeptree` yes this is useful, I use it. But solely with `--keeptree`, the mock struct and the original interface with the same name will end up in different paths, it's a little bit annoy when coding because you have to check if you have the right one imported. That's why I think it would be better if we could give it a prefix. It will be intuitive to refer to different things(struct/interface).

mockery/walker.go
```diff
@@ -123,7 +125,9 @@ func (this *GeneratorVisitor) VisitWalk(ctx context.Context, iface *Interface) e
 
 	if this.InPackage {
 		pkg = filepath.Dir(iface.FileName)
-	} else {
+	} else if this.PackageNamePrefix != "" {
+		pkg = fmt.Sprintf("%s%s", this.PackageNamePrefix, iface.Pkg.Name())
+	} else if this.PackageName != "" {
```
LandonTClipp, line 130: Could probably just do this:

```
if this.InPackage {
    pkg = filepath.Dir(iface.FileName)
} else {
    pkg = fmt.Sprintf("%s%s", this.PackageNamePrefix, iface.Pkg.Name())
}
```

mockery/walker.go
```diff
@@ -123,7 +125,9 @@ func (this *GeneratorVisitor) VisitWalk(ctx context.Context, iface *Interface) e
 
 	if this.InPackage {
 		pkg = filepath.Dir(iface.FileName)
-	} else {
+	} else if this.PackageNamePrefix != "" {
+		pkg = fmt.Sprintf("%s%s", this.PackageNamePrefix, iface.Pkg.Name())
```
LandonTClipp, line 129: What should the behavior be if `PackageNamePrefix` and `PackageName` are both specified? I think the way this is isn't super clear, because if you specify `PackageNamePrefix`, `PackageName` becomes ignored.

LandonTClipp:
@larryzhao if you could rebase off the most recent commit that would be great, it looks like there are some merge conflicts.

larryzhao:
> @larryzhao if you could rebase off the most recent commit that would be great, it looks like there are some merge conflicts.

Done rebased & conditions of precedence fixed :-D

larryzhao:
> > @larryzhao if you could rebase off the most recent commit that would be great, it looks like there are some merge conflicts.
> 
> Done rebased & conditions of precedence fixed :-D

LandonTClipp:
Thanks! It looks like you have a significant number of commits in here that aren't your own, you'll want to be sure to get rid of those.

larryzhao:
> Thanks! It looks like you have a significant number of commits in here that aren't your own, you'll want to be sure to get rid of those.
 
@LandonTClipp just no idea how it went this way. I just made new one here: https://github.com/vektra/mockery/pull/317 . this would be easier.

Issue #291, opened 2020-06-17
Invalid mock generated with the same package names


RossHammer:
Using either 1.1.2 or 2.0.0-alpha.13 release builds of mockery we have the following issue

`cloudformation/interface.go`
```go
package cloudformation

import "github.com/aws/aws-sdk-go/service/cloudformation"

type invalid interface {
	String(*cloudformation.CreateStackInput) string
}
```

`main.go`
```go
package main

func main() {}
```

generates a mock like like the following when using `mockery -all -testonly -inpkg`
```go
// Code generated by mockery v1.1.2. DO NOT EDIT.

package cloudformation

import mock "github.com/stretchr/testify/mock"

// mockInvalid is an autogenerated mock type for the invalid type
type mockInvalid struct {
	mock.Mock
}

// String provides a mock function with given fields: _a0
func (_m *mockInvalid) String(_a0 *cloudformation.CreateStackInput) string {
	ret := _m.Called(_a0)

	var r0 string
	if rf, ok := ret.Get(0).(func(*cloudformation.CreateStackInput) string); ok {
		r0 = rf(_a0)
	} else {
		r0 = ret.Get(0).(string)
	}

	return r0
}
```

You can see the import for cloudformation is missing and in turn the go reports undefined: cloudformation in the mock file.

It seems like when the cloudformation directory is at the root the issue cannot be re-created. When mockery is built from source either via `go get` or cloning and `go install` this issue seems to go away and the import is generated properly. As installing is not supported using `go get` anymore our mocks are no longer generating properly.
```go
package cloudformation

import (
	"github.com/aws/aws-sdk-go/service/cloudformation"
	mock "github.com/stretchr/testify/mock"
)
```

LandonTClipp:
What version of go are you using in `go get` ? I'd say this is a pretty serious issue, let's get this addressed.

LandonTClipp:
I can reproduce this issue as you describe. I'm looking for a possible cause.

LandonTClipp:
So I added some debug messages and discovered that mockery is skipping this explicitly:

```
[ltclipp@landon-virtualbox test_issue]$ ../mockery --inpackage --all --log-level debug
17 Jun 20 19:38 CDT INF Starting mockery dry-run=false version=0.0.0-dev
17 Jun 20 19:38 CDT INF Walking dry-run=false version=0.0.0-dev
[...]

PR #297, opened 2020-06-22
Fixing #291 where a different import with the same name as the package would not get included in the mock

RossHammer:
This should fix the issue from #291. The inconsistency there came from `FormatOnly` being set to false when re-formating the file because it is allowed to add/remove imports.

The logic for the test that was edited came from a test that was not quite right in #143 and in turn was enforcing that this bug existed. It was kind of testing what is was looking for but not correctly so I added a proper test to validate it. I could not make the test fail with the previous code though without putting a more full `pkg` path in the tests.


LandonTClipp:
Thanks for the PR, I'll get this reviewed within the next day or so and we can get it merged.

LandonTClipp:
(I haven't forgotten about this, will probably get to it on the weekend)

Issue #301, opened 2020-06-26
Generate mock for function type


georgysavva:
Imagine you want to abstract and mock something that can do only one thing. 
Instead of creating an interface with one method: 
```golang
type Sender interface {
    Send(data string) error
}
```
You could create a function type, because it's simpler in some cases:
```golang
type SendFunc func(data string) error
```
And then accept a function with `SendFunc` type, as you would do with the `Sender` interface. 
But in order to generate a mock, mockery needs an interface, it can't do it for a function, and you will need to write the mock manually. I think it's something that can be improved.
Example of a mock that could be generated for the `SendFunc` type in this case:
```golang
// MockSendFunc is an autogenerated mock type for the SendFunc type
type MockSendFunc struct {
	mock.Mock
}

// Execute provides a mock function with given fields: data
func (_m *MockSendFunc) Execute(data string) error {
	ret := _m.Called(data)

	var r0 error
	if rf, ok := ret.Get(0).(func(string) error); ok {
		r0 = rf(data)
	} else {
		r0 = ret.Error(0)
	}

	return r0
}
```
It has exactly 1 method with a general name like `Execute` or the name of the type without the `Func` suffix, i.e: `Send`.This method has the signature of the target function type. That the user will pass to the production code that expects a `SendFunc` type.
What do you think?

LandonTClipp:
I think the scope of the changes required to implement this would be large enough that I wouldn't want to attempt this. The entire code base is written with the assumption that you are mocking interfaces, not variables.

With that being said, if you can find a way to integrate it in with the existing code base without much trouble, shoot me a PR and we can discuss it.

georgysavva:
Okay, thanks for your response. I will try to approach this when I have free time.

PR #307, opened 2020-07-03
Generate mocks for function types

georgysavva:
Now mockery can generate mocks for function types. This feature was requested here: https://github.com/vektra/mockery/issues/301.

This is the simplest implementation that I could come up with. Please share your thoughts about it. I see a possible Improvement here:
1. Introduce a general interface: `TypeToMock`, that will replace the current `Interface` struct.
2. Create additional struct: `Function` 
3. Make both `Interface` and `Function` implement the `TypeToMock` interface

It's bit more complicated and maybe it's not worth it.


LandonTClipp:
Thanks for the great PR, I am going to be reviewing this over the next few days so I can fully understand the changes.

georgysavva:
Hi. Any update on this?

LandonTClipp:
Sorry for the delay! I looked at this a few days ago and don't see anything I'd want to change. I will go ahead and merge this and get it released.

PR #310, opened 2020-07-04
Remove 'docs' directory

nmiyake:
Directory contained empty README and a "Peek" gif that was not
referenced/used in any documentation.


LandonTClipp:
The gif is used in the main README so we can't remove this.

nmiyake:
Ah right just noticed that. I think having the name of the file be a bit more descriptive (something like `demo.gif` or `example.gif`) might help to make it clearer that this is an intended file, but not a big deal.

The README in this directory does seem empty, though -- opened #313 for this.

LandonTClipp:
Yeah I agree, if you want to submit a PR to make that name more useful that would be super appreciated, thanks!

PR #323, opened 2020-08-05
Speed up loading of source code

Helcaraxan:
The current approach of loading the source code of targeted packages has some inefficiencies that may induce considerable slowdowns and unnecessary lengthy runs when using `mockery` on larger codebases. Most notably it always defaults to loading individual files, even when specifying an entire package as target.

This PR speeds up the processing by rewiring the `Parse` and `Walk` methods to accept generic target patterns instead of only directory and file paths. This removes the need for the CLI tool to invoke `package.Load()` to determine the underlying files when given a package pattern and also means that all the source code is loaded in a single call to `package.Load()` instead of one call per source file in the targeted package(s).

A secondary advantage of this rewiring is that the `Walker` struct and it's API now also accept a package pattern in its `BaseDir` field making it more user friendly for users invoking the internals of `mockery` themselves instead of using the command-line tool as they benefit from the same speed-up (I am one of those).

There are no breaking changes in either the API or the semantics of the CLI tool and its implementation packages.

The PR also contains a few drive-by fixes:
- `Walker.doWalk` now properly reports errors. 
- The removal of an unused test utility function.
- The fixing of tests that were silently failing due to `Walker.doWalk` not properly reporting errors.
- The deprecation of the `Walker.LimitOne` field as it has no longer any speed-up benefit when set internally by the CLI side.

---

Some anecdotal testing on a large private codebase that generates 100s of mocks from ~150 packages by calling `Walker.Walk`  via internal tooling shows a runtime reduction of 65+% from 180 seconds down to 50-60.

### DIFF
--- a/pkg/parse.go
+++ b/pkg/parse.go
@@ -2,16 +2,16 @@ package pkg
 
 import (
 	"context"
-	"fmt"
 	"go/ast"
 	"go/types"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"sort"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
-	"github.com/vektra/mockery/v2/pkg/logging"
 	"golang.org/x/tools/go/packages"
 )
 
@@ -26,12 +26,14 @@ type Parser struct {
 	entries           []*parserEntry
 	entriesByFileName map[string]*parserEntry
 	parserPackages    []*types.Package
-	conf              packages.Config
+	conf              *packages.Config
 }
 
 func NewParser(buildTags []string) *Parser {
-	var conf packages.Config
-	conf.Mode = packages.LoadSyntax
+	conf := &packages.Config{
+		Mode:  packages.NeedFiles | packages.NeedImports | packages.NeedName | packages.NeedSyntax | packages.NeedTypes,
+		Tests: false,
+	}
 	if len(buildTags) > 0 {
 		conf.BuildFlags = []string{"-tags", strings.Join(buildTags, ",")}
 	}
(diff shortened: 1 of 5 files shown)


LandonTClipp:
Fantastic work, I will review this more in-depth over the coming days when I get some free time. Let's get this merged.

pkg/parse.go
```diff
@@ -42,67 +44,71 @@ func NewParser(buildTags []string) *Parser {
...
+	switch {
+	case os.IsNotExist(err):
+		// The pattern represents one or more packages and should be passed directly to the package loader.
+		log.Debug().Msgf("Loading packages corresponding to pattern %q.", pattern)
+		query = pattern
+
+	case !info.IsDir():
+		// A file should be passed directly to the package loader as a 'file' query.
+		if filepath.Ext(pattern) != ".go" {
+			return fmt.Errorf("specified file %q cannot be parsed as it is not a source file", pattern)
```
LandonTClipp, line 65: This should be `errors.Errorf` so that we can preserve the stack.

pkg/parse.go
```diff
@@ -42,67 +44,71 @@ func NewParser(buildTags []string) *Parser {
...
+		} else if strings.HasPrefix(info.Name(), ".") || strings.HasPrefix(info.Name(), "_") {
```
LandonTClipp, line 66: Is this not different behavior? I may be wrong but I dont' see anywhere where mockery would have skipped `.*` or `_*` files.

pkg/parse.go
```diff
@@ -42,67 +44,71 @@ func NewParser(buildTags []string) *Parser {
...
+	}
+
+	log.Debug().Msgf("parsing")
 
-		pkg := pkgs[0]
+	pkgs, err := packages.Load(p.conf, query)
+	if err != nil {
+		return err
+	} else if filepath.Ext(pattern) == ".go" && len(pkgs) > 1 {
+		err := fmt.Errorf("file %q maps to multiple packages (%d) instead of a single one", pattern, len(pkgs))
```
LandonTClipp, line 102: `errors.Errorf` here.

pkg/parse.go
```diff
@@ -42,67 +44,71 @@ func NewParser(buildTags []string) *Parser {
...
+		pattern, err = filepath.Abs(pattern)
+		if err != nil {
+			return err
 		}
 
-		pkgs, err := packages.Load(&p.conf, "file="+fpath)
+		log.Debug().Msgf("Loading file %q.", pattern)
+		query = "file=" + pattern
+
+	default:
```
LandonTClipp, line 78: We probably want this case to be `info.IsDir()` and then set the default case to error out (as it's an unknown condition).

pkg/parse.go
Helcaraxan: This is some [existing logic](https://github.com/vektra/mockery/blob/b4bca183629684bf40b069cc339ebc22015ded71/pkg/walker.go#L75-L77) that I took from `Walker` and moved here. I am not partial to keeping or removing it, I just moved to keep the existing semantics. :slightly_smiling_face: 

Technically it would be a breaking change to remove this.

Helcaraxan:
Quick and friendly ping on this @LandonTClipp. No worries if you are too busy as I very much understand that other things may take priority. Just did not want it to slip of your radar. :slightly_smiling_face:

LandonTClipp:
Thanks for the ping, I will not let it slip! Life got in the way (car transmission got totaled, big fun). Also extra busy this week but I will be sure to come back.

LandonTClipp:
There have been other places where people suggested moving towards a package-specify model instead of file-specify model, and I'm in total agreement. It's much better to leverage the code already given in `package.Load`. You are right, however, that it would be a breaking change, which is why we need to wait for v3.

https://github.com/vektra/mockery/projects/3

I'm still in the process of collecting requirements for v3, and also want to finish/incorporate my other project into v3 https://github.com/chigopher/pathlib.

LandonTClipp:
Unfortunately it looks like this has a major bug in it:

```
12 Aug 20 22:07 CDT DBG Loading files in directory ".git/objects/0f". dry-run=false version=0.0.0-dev
12 Aug 20 22:07 CDT DBG parsing dry-run=false version=0.0.0-dev
^C
[ltclipp@landon-virtualbox pathlib]$ 
```

It wants to iterate over dot-files, so we have a regression.

PR #326, opened 2020-08-13
Revert "Speed up loading of source code"

LandonTClipp:
Reverts vektra/mockery#323


Helcaraxan:
Ah. Just realised this got reverted. Let me have a look at if I can add a regression test for this and fix the PR.

PR #329, opened 2020-08-17
Speed up loading of source code - v2

Helcaraxan:
Second edition of #323, now improved with a regression test for the walking of excluded directories (the ones with a name starting with `.` or `_`).

This contains the changes from the previous PR but with improved ignore logic that now exists both in `Parse` but also again in `Walker` where it already existed.

The added regression tests actually fails with the logic from the original PR but passes on this improved version.

### DIFF
--- a/cmd/mockery.go
+++ b/cmd/mockery.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"path/filepath"
 	"regexp"
 	"runtime/pprof"
 	"strings"
@@ -20,7 +19,6 @@ import (
 	"github.com/vektra/mockery/v2/pkg/config"
 	"github.com/vektra/mockery/v2/pkg/logging"
 	"golang.org/x/crypto/ssh/terminal"
-	"golang.org/x/tools/go/packages"
 )
 
 var (
@@ -50,7 +48,6 @@ func printStackTrace(e error) {
 			fmt.Printf("%+s:%d\n", f, f)
 		}
 	}
-
 }
 
 // Execute executes the cobra CLI workflow
@@ -86,7 +83,7 @@ func init() {
 	pFlags.String("filename", "", "name of generated file (only works with -name and no regex)")
 	pFlags.String("structname", "", "name of generated struct (only works with -name and no regex)")
 	pFlags.String("log-level", "info", "Level of logging")
-	pFlags.String("srcpkg", "", "source pkg to search for interfaces")
+	pFlags.String("srcpkg", "", "source package(s) to search for interfaces, may be a single package name or a package pattern (example: 'github.com/mockery/vektra/...'")
 	pFlags.BoolP("dry-run", "d", false, "Do a dry run, don't modify any files")
 
 	viper.BindPFlags(pFlags)
@@ -136,7 +133,6 @@ func (r *RootApp) Run() error {
 	var recursive bool
 	var filter *regexp.Regexp
 	var err error
-	var limitOne bool
 
 	if r.Quiet {
 		// if "quiet" flag is set, disable logging
@@ -171,7 +167,6 @@ func (r *RootApp) Run() error {
 			}
 		} else {
 			filter = regexp.MustCompile(fmt.Sprintf("^%s$", r.Config.Name))
-			limitOne = true
 		}
 	} else if r.Config.All {
 		recursive = true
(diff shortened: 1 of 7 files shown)


LandonTClipp:
Bah this doesn't seem to work either. Let me take a closer look at your code.

```go
[ltclipp@landon-virtualbox mockery]$ ./mockery --log-level debug
Using config file: /home/ltclipp/git/vektra/mockery/.mockery.yaml
19 Aug 20 20:58 CDT INF Starting mockery dry-run=false version=0.0.0-dev
19 Aug 20 20:58 CDT INF Walking dry-run=false version=0.0.0-dev
19 Aug 20 20:58 CDT DBG Not loading path "." as it is prefixed with either '.' or '_'. dry-run=false version=0.0.0-dev
19 Aug 20 20:58 CDT DBG Skipping path "." as it is prefixed with either '.' or '_'. dry-run=false version=0.0.0-dev
```

LandonTClipp:
So here's the problem I have found with this PR:

1. The current PR doesn't work because if pattern refers to a filesystem path, it is not made absolute, thus it excludes paths like `.` from parsing, thus it parses nothing. The code previously made every path absolute and then essentially did a `filepath.Base()` on it and performed the logic using that.
2. It seems that the current runtime using the old-fashioned "pattern=filesystem path" is now 5 times as long. This is a simple bug fix that I added.

With the fixes from this PR implemented:
```
real    0m15.607s
user    0m22.111s
sys     0m11.066s
[ltclipp@landon-virtualbox mockery]$ 
```

v2.2.1:

```
real    0m4.141s
user    0m5.689s
sys     0m3.274s
[ltclipp@landon-virtualbox mockery]$ 
```

I am going to work from my side to try to figure out why this is slower and if there is a reasonable fix for it.

tuananh:
@LandonTClipp 
looks like it's 4 time slower from your test. btw it would be nice to have a benchmark added to CI for every merge request to detect perf regression.

LandonTClipp:
@tuananh oops yes you are right, a previous run had the difference at 5 but the run I actually posted is ~4.

LandonTClipp:
So I am thinking about it a bit more and I am leaning towards not merging this. I've taken a deeper dive into what this functionality actually involves (admittedly, I didn't look at it too closely the first time around), and it is edging too close to what v3 is trying to accomplish. 

The spirit of this PR is really really exciting and I really love the direction that we are trying to push here. I'm jotting down notes for v3. However I don't feel it's appropriate to be placing it here just yet. It's not the right time.

In v3, we are going to move towards a model where you specify the qualified package name only, and not iterate over files (they are two distinctly different modes). It will be a configuration-file-driven thing and will cut down immensely on runtime, as it loads per-package instead of per-file like you said in your original PR. It will also cut down a huge amount of complexity.

So long story short, I think it will be worth it for us to wait for me to do the v3 rewrite. I am going to be removing lots of cruft that has accumulated over the years. It's going to be really nice, but will take a bit of time. 

Thoughts?

Helcaraxan:
I am really sorry about this again not fixing the path-based specifications. It highlights the need for more tests (not a criticism, just an observation) but if you are indeed going for v3 those path-based tests might not be worth writing.

At the higher-level I am, from my own biased perspective, keen on merging this as it should not constitute a breaking change (once the path issue is fixed and any observed slowdown is addressed). However I also very much understand your point about a proper v3 being maybe the preferable path to go down. My own needs could, for now, be addressed by instead using my fork with this change as I only use qualified package names. So if you don't feel like merging it even with potential fixes I understand.

Regarding the wider v3 rewrite, I'd be happy to provide manpower to help you out. It does however require some level of guidance or direction from you, as well as review time. I currently can't find a good clear picture of what a v3 should entail as clean-ups and improvements. A starting point might be to create a `v3` issue-label and mark relevant GitHub issues with it. That and add more issues about the more technical implementation details that you'd like to change and not just the modified features like required qualified package names.

LandonTClipp:
No worries @Helcaraxan. I picked up this project a few months ago after a long period of inactivity so all of your observations are totally and 100% valid :) I like the project enough that I'm dedicating my time to improving it so much that it becomes the de-facto standard for go mocking. I have a few other lofty goals for this project that I have not even mentioned.

> So if you don't feel like merging it even with potential fixes I understand.

I would say, keep developing your branch and push them here. I am eager to see how it evolves. If the changes are robust enough and the behavior can be proven to be backwards-compatible, I might be open to merging it. But I've declared a general change freeze to v2 due to a continual increase in complexity of the code base, which I want to avoid. Most of this complexity will be eliminated in v3.

> Regarding the wider v3 rewrite, I'd be happy to provide manpower to help you out.

I would definitely enjoy more help. I think a serialization point right now is getting the structure of the code cemented. Once I can refactor it enough where more contributors can add, I would love to leverage whatever time/energy you can spend.

So in general, please keep up the great work and let's see what you can come up with in your PR!

LandonTClipp:
Closing this PR as we probably won't be merging this. Thanks for the thoughts/contributions though.

PR #339, opened 2020-10-15
Idiomatic Go

kensipe:
More idiomatically Go

1. dedented else statement which could only be true
2. removed references to `this` and removing receiver handles where handles are not used 
3. enforced consistency of receiver handles

Signed-off-by: Ken Sipe <kensipe@gmail.com>


LandonTClipp:
Thanks for the PR, looks good to me!

PR #344, opened 2020-11-18
Generate Constants for Methodnames

tbaehler:
Hi all

First thanks for your great work !
Maybe a feature that I miss, would also fit for others. So I created this PullRequest
The Idea is to generate constants for method names. 
See the example below 

	 func TestIt() {
	       cleanupMock := mocks.Cleanup{}
	       //do not use strings
	       cleanupMock.On("Execute"). Return()
	       //use generated constants
	       cleanupMock.On(mocks.Cleanup_Execute). Return()
	}

This makes it easier to refactor the code. 

See generator.go / generator_test.go and walker.go for my changes.


LandonTClipp:
Hi @tbaehler, thanks for the PR! I will leave this up for consideration for probably a week or so as a matter of practice. The PR itself makes good sense and I think it could provide a benefit to folks.

LandonTClipp:
@tbaehler this PR does not seem to have a way to handle collisions in these variable names. It's possible this might introduce breaking changes for some folks.

tbaehler:
@LandonTClipp 

Can you provide an example?

Problem one: Without my change

If you have two packages with the same Interface 
```go
package a  
  
type A interface {  
   hello()  
}

package b  
  
type A interface {  
   world()  
}
```

If you generate the mocks in the same directory, then the last definition wins. You will never see a generated Mock with method hello(). 
Thats the way it works today. 

If you generate in different directory. 
You will end up with a conflict because Interface mock.A exists twice. 

The constant name is generated like so: <interface_name><function_name> and mock.interface_name must be unique. 

So the only way to have a conflict is, that the interface already defines a constant with exactly this name. This is possible but very unlikely. 
Or do you see any other conflicts?

LandonTClipp:
```
So the only way to have a conflict is, that the interface already defines a constant with exactly this name. This is possible but very unlikely.
```
Well when dealing with a global audience, the probability becomes "likely." If it breaks for even a single person, we can't add it without a major version bump. It's something I would love to consider for v3 but I don't think it's appropriate in this case.

I added this to the TODO for v3: https://github.com/vektra/mockery/projects/3, hopefully that is an acceptable compromise?

LandonTClipp:
I am closing this as we now have type safe mocks through the expecter interface. Thanks folks.

PR #351, opened 2020-11-27
Add --boilerplate-file flag

### LINKED ISSUE #350: Support adding boilerplate to generated files
Some projects may require files to begin with a boilerplate comment, e.g. a license. The CLI should be able to support reading a boilerplate file or text and prepending it to each generated file.

avorima:
Flag should point to a file containing a go block comment. Its content is written at the start of each generated file.

Fixes #350


LandonTClipp:
Thanks for the PR @avorima. This is approved.

pkg/generator.go
```diff
@@ -265,6 +265,13 @@ func (g *Generator) GeneratePrologueNote(note string) {
 	}
 	prologue += ". DO NOT EDIT.\n"
 
+	// The note contains a boilerplate block comment that needs to come first.
+	if strings.HasPrefix(note, "/*") {
+		g.printf("%s\n", note)
+		g.printf("\n")
+		note = ""
+	}
```
LandonTClipp, line 273: I think this is slightly inconsistent and non-obvious behavior. The location of the note is reliant on whether or not it has a block comment. I think that there should be explicit parameters that define whether or not it appears at the beginning.

The other thing that is inconsistent is that `//` is appended to the note in normal circumstances, but if the note has `/*` then its not prefixed. I would recommend instead saying something like "if note has a newline, use block comments."

LandonTClipp (changes requested):

pkg/generator.go
avorima: Yes, sorry about that. I just tried to make it work for my use-case which is adding a boilerplate comment at the beginning of the file. I think I'll just leave the current note code as is and expose that boilerplate functionality via `--boilerplate` or `--boilerplate-file` flag. WDYT?

LandonTClipp: I like the idea of `--boilerplate-file`, I think that will maintain a lot of backwards-compatibility. I'll review this in the next day or so!

LandonTClipp (approved):
Looks good, will get this merged in a day or so.

PR #357, opened 2021-01-06
Add --exported flag

### LINKED ISSUE #191: Add option to generate exported mock type for internal interfaces

mneverov:
Add --exported flag to generate public mocks from private interfaces.

Fixes #191 

@LandonTClipp could you please take a look?


pkg/generator_test.go
```diff
@@ -246,6 +246,39 @@ func (_m *mockRequester_unexported) Get() {
...
+type MockRequester_unexported struct {
+	mock.Mock
+}
+
+// Get provides a mock function with given fields:
+func (_m *MockRequester_unexported) Get() {
+	_m.Called()
+}
+`
+	generator := NewGenerator(
```
mneverov, line 260: Note, that `checkGeneration` accepts some of the configs, but not all. Instead of introducing another boolean argument for the function I decided to create a generator manually.

pkg/config/config.go
```diff
@@ -12,6 +12,7 @@ type Config struct {
 	Dir                  string
 	DisableVersionString bool `mapstructure:"disable-version-string"`
 	DryRun               bool `mapstructure:"dry-run"`
+	Exported             bool
```
LandonTClipp, line 15: Just for consistency let's add the explicit mapstructure tag here (really I would like all of them to have it).

pkg/generator.go
```diff
@@ -201,6 +201,9 @@ func (g *Generator) mockName() string {
 		if ast.IsExported(g.iface.Name) {
 			return "Mock" + g.iface.Name
 		}
+		if !ast.IsExported(g.iface.Name) && g.Exported {
```
LandonTClipp, line 204: Better yet, let's just say `if g.Exported { }`.

LandonTClipp (changes requested):
Some comments, otherwise looks good. Address those and I can approve.

pkg/config/config.go
mneverov: fixed

pkg/generator.go
mneverov: you're right, fixed.

LandonTClipp (approved):
Approved, will leave up for a few days before merging per procedure.

pkg/generator.go
```diff
@@ -201,6 +201,9 @@ func (g *Generator) mockName() string {
 		if ast.IsExported(g.iface.Name) {
 			return "Mock" + g.iface.Name
 		}
+		if g.Exported {
+			return "Mock" + strings.Title(g.iface.Name)
```
LandonTClipp, line 205: Actually after looking at this again, would it make more sense to instead just do this instead?

```go
if g.Exported or ast.IsExported(g.iface.Name) {
	return "Mock" + g.iface.Name
}
```
The casing of the interface name doesn't matter since we prepend with `Mock` which makes it exported. Less logic would be better. What is your opinion?

mneverov: Agree, updated the PR.

LandonTClipp: Great, once PR builder is done I'll merge.

PR #359, opened 2021-01-09
Add --vararg flag

mneverov:
Currently, mockery unrolls variadic arguments and passes them as a separate arguments, see interface [RequesterVariadic](https://github.com/vektra/mockery/blob/e725139b5f9a25fcbb2c40a43b7d16c64b3edfbd/pkg/fixtures/requester_variadic.go#L5) and corresponding [mock](https://github.com/vektra/mockery/blob/e725139b5f9a25fcbb2c40a43b7d16c64b3edfbd/mocks/pkg/fixtures/RequesterVariadic.go#L24).
According to the [comment in the generator](https://github.com/vektra/mockery/blob/e725139b5f9a25fcbb2c40a43b7d16c64b3edfbd/pkg/generator.go#L598) it was done to `avoid incompatibility issues with the diff algorithm in github.com/stretchr/testify/mock.`. I believe testify did have some issues with varargs and they were fixed.
With the current approach it is impossible to use `mock.Anything` as a substitution for a variadic argument. If a function under test received a vararg of 15 distinct values, one needs to pass 15 `mock.Anything` to the generated mock.

This PR introduces a new `vararg` flag (please suggest a better name). When mockery is run with `--vararg` flag mocks for the functions with variadic arguments will be generated as is, i.e. vararg is a single argument. This allows to use `mock.Anything` for such arguments.

@LandonTClipp could you please take a look?


pkg/generator_test.go
```diff
@@ -702,6 +707,127 @@ func (s *GeneratorSuite) TestGeneratorVariadicArgs() {
...
+	expectedLines := strings.Split(expected, "\n")
+	actualLines := strings.Split(string(actual), "\n")
+
+	s.Equal(
+		expectedLines, actualLines,
+		"The generator produced unexpected output.",
+	)
+}
+
+func TestRequesterVariadicOneArgument(t *testing.T) {
```
mneverov, line 741: I've added tests to check if testify works correctly with the new mocks. Please let me know if I better delete them.

pkg/generator_test.go
```diff
@@ -702,6 +707,127 @@ func (s *GeneratorSuite) TestGeneratorVariadicArgs() {
...
+func (s *GeneratorSuite) TestGeneratorVariadicArgsAsOneArg() {
+	// Read the expected output from a "golden" file that we can also import in CompatSuite
+	// to asserts its actual behavior.
+	//
+	// To regenerate the golden file: make fixture
+	expectedBytes, err := ioutil.ReadFile(filepath.Join(fixturePath, "mocks", "requester_variadic_one_arg.go"))
+	s.NoError(err)
+	expected := string(expectedBytes)
+	expected = expected[strings.Index(expected, "// RequesterVariadicOneArgument is"):]
+	generator := NewGenerator(
```
mneverov, line 719: I created a new generator manually again as it needs new `Vararg` flag.
Does it make sense to make `checkGeneration` accept config? maybe in a different PR, or does it better suit mockery `v3`?

LandonTClipp:
Thanks for the PR @mneverov. I understand the issue here and I think this is a reasonable solution. I think the argument name may need to be changed... let me think about this. Otherwise I like the solution.

cmd/mockery.go
```diff
@@ -91,6 +91,7 @@ func init() {
 	pFlags.BoolP("dry-run", "d", false, "Do a dry run, don't modify any files")
 	pFlags.Bool("disable-version-string", false, "Do not insert the version string into the generated mock file.")
 	pFlags.String("boilerplate-file", "", "File to read a boilerplate text from. Text should be a go block comment, i.e. /* ... */")
+	pFlags.Bool("vararg", false, "Generates mocks with vararg as a single argument.")
```
LandonTClipp, line 94: I think better terminology here would be to say "For functions with variadic arguments, do not unroll the arguments into the underlying testify call. Instead, pass variadic slice as-is."

As for the name, I think it makes more sense to call if `--no-unroll-variadic`. It's more verbose but is clearer what it does. Thoughts?

mneverov: I like `--unroll-variadic`. Cons: it will be used in the project with negation only `if !g.UnrollVariadic ...`, which I think is reasonable cost for the self-explanatory argument.
If you are ok with it I'll make the adjustments.

LandonTClipp: Yeah let's go ahead with that, I think it will make more sense. Thanks for the time you're putting into this!

pkg/generator_test.go
LandonTClipp: I think making a new generator is the right way to do it. Generator objects should be tied to static config. I have no problems with this.

cmd/mockery.go
mneverov: thx for the project!

Issue #306, opened 2020-07-01
Officially support go get


kollektiv:
The README mentions `go get` as officially deprecated. However any Go project using Go modules naturally supports `go get`

Could `go get` be officially supported since it already works, and is no extra effort for the maintainers? This will also support the [tools.go](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) method.

README would need to be updated to use `go get github.com/vektra/mockery/v2@semver`

https://github.com/vektra/mockery/issues/299 has already resolved `go get`ing v2 of mockery

LandonTClipp:
One issue we'd need to figure out is if we're going to officially support `go get`, then we need to figure out how the semver will be consistent with built versions of the package. The correct version is injected build-time, so `go get` will only give you a `v0.0.0-dev` which isn't useful.

kollektiv:
@LandonTClipp [`debug.ReadBuildInfo().Main.Version`](https://golang.org/pkg/runtime/debug/#ReadBuildInfo) might be an option. I believe it populates with a semver version on `go get`

LandonTClipp:
@kollektiv could you confirm this? The only way I can see that'd work is if it's getting the semver from the git tag. Otherwise, how would it get it?

kollektiv:
@LandonTClipp The go get tooling knows how to get the semver from git tags (or other version control tags). It has to be able to do this to populate go.mod with versions. There's a lot more info about version control support at https://golang.org/ref/mod#tmp_41

I've created a proof of concept at https://github.com/kollektiv/examplegotool to confirm that this method works

```bash
# go get is ran outside of GOPATH and outside a go module project. go get will install to $GOBIN, and $GOBIN is on my $PATH
$ GO111MODULE=on go get github.com/kollektiv/examplegotool
go: github.com/kollektiv/examplegotool upgrade => v1.0.0

$ examplegotool
2020/07/06 22:05:18 main module version: v1.0.0

$ GO111MODULE=on go get github.com/kollektiv/examplegotool@e73dabe299e9ae80f7f01b16d073b5b1fea5755d
go: downloading github.com/kollektiv/examplegotool v1.0.1-0.20200707050442-e73dabe299e9
go: github.com/kollektiv/examplegotool e73dabe299e9ae80f7f01b16d073b5b1fea5755d => v1.0.1-0.20200707050442-e73dabe299e9

$ examplegotool
[...]

PR #363, opened 2021-02-10
Pull Semver version from the buildinfo

spencercornish-wk:
As per the discussion in #306, it should be possible to use Go's build info to get version information, as opposed to injecting it in during build time. This has the benefit of making version info populate for `go get` installs.

This PR attempts to do that. It also falls back to using the injected version info if the `BuildInfo` is not accessible for some reason.

P.S. Local builds will now have the tag `(devel)`, as that is the default from the `BuildInfo`

fyi @kollektiv @LandonTClipp


LandonTClipp:
Sorry for the delay, this is still on my radar. Will get to it soon.

LandonTClipp:
This appears to have not worked unfortunately. Downloading the pre-built binary just does this:

```
[ltclipp@landon-virtualbox tmp]$ ./mockery --version
17 Mar 21 22:05 CDT INF Starting mockery dry-run=false version=(devel)
(devel)

```

We'll have to revert.

Issue #328, opened 2020-08-17
is there any tips to speedup mockery generation?


tuananh:
I test on a reasonably small project and it's taking over a minute. is there anyway to speed it up?

I use `mockery --all --keeptree` command

```
________________________________________________________
Executed in   75.11 secs   fish           external
   usr time  228.08 secs  389.00 micros  228.08 secs
   sys time   58.53 secs  342.00 micros   58.53 secs
```

LandonTClipp:
There was a PR that attempted to help fix this: https://github.com/vektra/mockery/pull/323. Unfortunately there is not any  advice I can give at the moment.

We tried merging it but it introduced a regression. Hopefully @Helcaraxan can take another look at it and see if there's any way we can fix it.

There is a rather large overhaul that this codebase needs to undergo, which is scheduled for v3. It has accumulated a significant amount of bloat, and was also made with pre-golang-modules assumptions. 

One of the major tasks that needs to be accomplished is switching from walking/loading individual files, to loading packages. This is discussed in the PR, and has also been brought up in various other places.

LandonTClipp:
@tuananh I have a potential solution to this now, view the linked draft PR.

PR #371, opened 2021-03-12
Add package load cache

lionello:
In our codebase Mockery `v2.6.0` takes about 50 seconds per generated mock. Using `--log-level=trace` revealed that files were getting parsed multiple times: for each file in a folder, `packages.Load` would get called for all files in that folder. In addition to this O(N^2) behavior, when using `--inpackage` Mockery would also parse the previously generated mocks, causing another big slow down for folders with many mocks in them.

The bad performance was successfully mitigated by wrapping the call to `packages.Load()` with a simple cache and skipping files starting with `mock_` (much like the existing filter for `_test.go`). This cut down the runtime of Mockery on our repository from ~40 minutes to ~10 minutes, a ~75% improvement!

### DIFF
--- a/pkg/parse.go
+++ b/pkg/parse.go
@@ -22,11 +22,17 @@ type parserEntry struct {
 	interfaces []string
 }
 
+type packageLoadEntry struct {
+	pkgs []*packages.Package
+	err  error
+}
+
 type Parser struct {
 	entries           []*parserEntry
 	entriesByFileName map[string]*parserEntry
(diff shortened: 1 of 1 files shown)


lionello:
This might also address #328

LandonTClipp:
This looks great. The only thing is that not everyone's project will use the `mock_` prefix to define their mocks. We probably need a more robust way to determine which files are mocks and which aren't. That isn't a blocker for this specific PR but it's something to note.

Issue #367, opened 2021-03-02
--exported flag not working


karlc1:
I downloaded mockery today using `go get github.com/vektra/mockery/v2/.../`, so I cannot check exact version number, but `--exported` is documented in `mockery --help` output. Anyway, when I run `mockery --exported --all`, none of my unexported interfaces are generated as exported mocks. Instead, the generated code looks exactly the same as if I omit the --exported flag. Am I doing something wrong or is this broken?

LandonTClipp:
It's possible it's not working correctly, if you can find where in the code it's causing this behavior, we can discuss in more specific detail. PRs are always welcome!

ijalalfrz:
@LandonTClipp i have the same issue my interface is private:

```
type elasticsearchRequest interface {
	Do(ctx context.Context, transport esapi.Transport) (resp *esapi.Response, err error)
}
```
then i did ```mockery --exported --name=elasticsearchRequest```
i try to mock it with mockery v2.7 but the exported interface is still in private

LandonTClipp:
We are going to be deprecating the non-`packages` config semantics pretty soon, so I'm going to mark this as "wontfix". The feature you're looking for is supported in `packages`.

PR #379, opened 2021-04-09
implement exported flag also when not InPackage

k15r:
`exported` did not work when mockery was working not in InPackage mode. With this PR it also allows to export unexported interfaces in the other mode.

See #367


k15r:
Im not sure about the naming of exported interfaces in InPackage mode though:
`requested_unexported` becomes `Mockrequested_unexported`
whereas unexported interfaces look like this:
`mockRequested_unexported`

Would it not make sense to rename in the first case to `MockRequested_unexported`

LandonTClipp:
This looks fine, I don't see anything I'd want to change. I'll merge this. I have a suspicion that this might break something somewhere but if it does we can roll back. I'll get it pushed in the next day or so.

remychantenay:
Could we push this through? @LandonTClipp 🙏

PR #387, opened 2021-05-25
Added ResetMocks to the generator to be added to all generated mocks

joshualawson:
This allows us to `ResetMocks` easily when writing unit tests, as sometimes
mocks cause unwanted issues in certain scenarios, such as table tests,
where we want to reset the count of calls or change functionality of
calls etc.

This method is added to the end of generated mocks as a pointer receiver, it simply sets the mocks `Calls` and  `ExpectedCalls` back to their initial state.

Not sure this is the right approach to adding this, so feedback and suggestions on making this better are more then welcome!

### DIFF
--- a/pkg/fixtures/mocks/requester_variadic.go
+++ b/pkg/fixtures/mocks/requester_variadic.go
@@ -86,3 +86,9 @@ func (_m *RequesterVariadic) Sprintf(format string, a ...interface{}) string {
 
 	return r0
 }
+
+// ResetMocks resets the Calls and ExpectedCalls
+func (_m *RequesterVariadic) ResetMocks() {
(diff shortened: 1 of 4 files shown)


LandonTClipp:
What's the behavior if the original interface has a method called `ResetMocks()`? You'll probably want to account for that somehow.

Gevrai:
I feel like this is a PR that should be a function on [testify](https://github.com/stretchr/testify/)'s `mock.Call` directly, it isn't really related to mockery.

joshualawson:
@Gevrai after some thought, I felt the same, this should be part of testify's mocking library over being generated here. I think this PR can be closed

PR #392, opened 2021-06-24
Add Vanilla Mock Generator

aljorhythm:
Intention is to generate golang vanilla style mocks. I've piggy backed this repo because there's alot of reusable logic. Would it be better to create a new repo instead? If not would like advice on how to integrate these changes consistently and properly with existing source. `Generator` is a struct instead of an `interface`, so it might get messy.


aljorhythm:
writing in a separate repo: https://github.com/aljorhythm/vanilla-mock

Issue #397, opened 2021-07-21
v2.9.0 has wrong version string


MyChaOS87:
```
# go install github.com/vektra/mockery/v2@v2.9.0
# mockery --version                             
21 Jul 21 14:14 CEST INF Starting mockery dry-run=false version=v0.0.0-dev
v0.0.0-dev
```

version reported is v0.0.0-dev
same is put up in generated files

```
// Code generated by mockery v0.0.0-dev. DO NOT EDIT.
```

williambanfield:
I've noticed this as well. It looks like the config package defines a default version string, here: https://github.com/vektra/mockery/blob/master/pkg/config/config.go#L4.

This version variable is then filled in in the `goreleaser` step here:
https://github.com/vektra/mockery/blob/master/.goreleaser.yml#L9

The result is that when getting mockery using `go get`, the version will not be filled in. Looks like this was fixed previously:
https://github.com/vektra/mockery/issues/362
Perhaps the fix this is the same?

PR #398, opened 2021-07-30
version string logic updated to use v0.0.0-dev when no string set

williambanfield:
This change is aimed at addressing issues where mockery will use the incorrect version if `SemVer` variable is not set during build using the `-ldflag` option.

When running with ld flag set:
```
go build -ldflags="-X 'github.com/vektra/mockery/v2/pkg/config.SemVer=v1.2.3'"
[Fri 12:27] mockery[wb/397]:. ◯ ./mockery --version
Using config file: /home/william/repos/mockery/.mockery.yaml
30 Jul 21 12:27 EDT INF Starting mockery dry-run=false version=v1.2.3
v1.2.3
```

When running as module:
```
[Fri 12:40] ~/repos ◯ mkdir testmodule
[Fri 12:40] ~/repos ◯ cd testmodule
[Fri 12:40] ~/repos/testmodule ◯ go mod init github.com/testmodule
go: creating new go.mod: module github.com/testmodule
[Fri 12:40] ~/repos/testmodule ◯ go get github.com/vektra/mockery/v2/.../@v2.9.0
go get: added github.com/vektra/mockery/v2 v2.9.0
[Fri 12:41] ~/repos/testmodule ◯ go mod edit -replace github.com/vektra/mockery/v2@v2.9.0=../mockery
[Fri 12:41] ~/repos/testmodule ◯ go run github.com/vektra/mockery/v2 --version
30 Jul 21 12:41 EDT INF Starting mockery dry-run=false version=v2.9.0
v2.9.0
```

When running otherwise 
```
go run main.go --version
Using config file: /home/william/repos/mockery/.mockery.yaml
30 Jul 21 12:24 EDT INF Starting mockery dry-run=false version=v0.0.0-dev
v0.0.0-dev
```

closes: #397

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,10 +1,23 @@
 package config
 
+import "runtime/debug"
+
+const (
+	_defaultSemVer = "v0.0.0-dev"
+)
+
 // SemVer is the version of mockery at build time.
-var SemVer = "v0.0.0-dev"
+var SemVer = ""
 
 func GetSemverInfo() string {
-    return SemVer
+	if SemVer != "" {
+		return SemVer
+	}
+	version, ok := debug.ReadBuildInfo()
+	if ok && version.Main.Version != "(devel)" {
+		return version.Main.Version
+	}
+	return _defaultSemVer
 }
 
 type Config struct {


LandonTClipp:
This will probably work. I'll test tonight and see if it fixes it. I'm tempted to move towards hard coding the semver variable to the actual version, then the build pipeline would key off that to create the git tag. But, if this works we can go with it.

williambanfield:
Thanks so much for taking a look @LandonTClipp ! Let me know if there's anything I can do to support. Both solutions seem pretty good.

LandonTClipp:
I actually think I'm going to go with hard-coding the version string and having that be the source of truth, instead of relying on two disjointed methods (i.e. variable injection for builds, and ReadDebugInfo for modules). I've been getting pretty frustrated with this issue and it would solve it once and for all. The trick is getting the build pipelines to correctly key off of it and tag as appropriate.

LandonTClipp:
Actually, I changed my mind. I'll give this a shot. Might be easier.

PR #402, opened 2021-09-01
fix: Avoid package name collision with inPackage (Fixes #291)

### LINKED ISSUE #291: Invalid mock generated with the same package names

rws-github:


LandonTClipp:
@rws-github can you please add an empty commit to trigger the actions builder (just to verify tests still pass).

LandonTClipp:
retest this please

LandonTClipp:
Hi there, please close this PR and submit a new one. I just got github actions working with the help of another contributor and would like your changes to run through it. Thanks!

Issue #394, opened 2021-06-29
Latest release lacks compiled binaries


heyvito:
Hi there! I have some scripts that automatically downloads mockery, but it failed with version 2.9.0, which for some reason lacks precompiled binaries.

Just a heads-up in case that was unintended.

Cheers!

Edit: Version 2.8.0 reports itself as `version=(devel)`

LandonTClipp:
Hi, I'm aware of the issue. Our entire build pipeline broke when travis-ci was shut down. I have to port it over to github actions. I'm not sure when I can do this as I'll be pretty busy for the next few weeks. If anyone wants to pitch in that'd be great.

PR #403, opened 2021-09-09
Migrate to GitHub Actions

xakep666:
This PR adds 2 workflows for Actions:
* Go Test. It runs tests (only unit-tests currently) and uploads coverage report to Codecov if tests runs successfully. This workflow runs on pull request with target branch 'master' and push to branch 'master'.
* Release. It runs `goreleaser` with preparations (docker login) for each new tag.

Following secrets must be added:
* `DOCKER_USERNAME`, `DOCKER_PASSWORD` - DockerHub credentials
* `HOMEBREW_TAP_TOKEN` - Token for Homebrew tap.
* `CODECOV_TOKEN` - Token for Codecov

This PR should solve problems like #394

Also this PR removes unused CI configurations: `travis-ci` and `appveyor`.


xakep666:
@LandonTClipp you mentioned github actions port in #394 so probably you can review this.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
grongor, line 11: I think Windows is still supported, so you should include it here too. Maybe including macOS isn't a bad idea either.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
```
grongor, line 18: You should also include the "tip" version, as it was before (the latest Go version)

grongor (changes requested):

.github/workflows/testing.yml
xakep666: Looks good but may lead to bloat: (Linux, Windows, Mac)×(1.16,tip)
Splitting to different parallel workflows may help, will try this

grongor: I think you can use matrix builds for that. Here is an example how I did it with my library and I think it works well: https://github.com/grongor/panicwatch/blob/master/.github/workflows/ci.yaml

xakep666: Thanks, didn'k know about matrix feature in Actions

LandonTClipp:
This is great, thanks for taking the time. I will look at this tonight and approve if it looks good.

LandonTClipp:
It appears I don't have permissions to modify secrets on this repo, I need to contact the owner so I can get added.

LandonTClipp:
I was able to access secrets through the API. With some minor changes, the build worked wonderfully. Homebrew, dockerhub, GH release page all working. Thanks a ton :1st_place_medal:

LandonTClipp:
We'll want to implement the testing matrix for different version but in the immediate term this gets us out the door.

PR #399, opened 2021-08-18
Makefile fix, and tests refactoring

grongor:
I'm not sure if mocks like `mocks/cmd/stackTracer.go ` were being generated for a reason, if so then I can add them back. However, they were not used anywhere so I went ahead and deleted them.

PR #406, opened 2021-09-13
Add constructor for mocks

grongor:
These changes will simplify the tests by not being required to assert mocks expectations every time. This will also eliminate situations where your tests are not failing even though they should, only to find out that you forget the defer statement.

Example
Before:
```go
factory := &mocks.Factory{}
defer factory.AssertExpectations()

something := &mocks.Something{}
defer something.AssertExpectations()

something.On("GetTime").Return("tooLate")

factory.On("Create").Return(something)

testThings(factory)
```
Now:
```go
factory := mocks.NewFactory()
something := mocks.NewSomething()

something.On("GetTime").Return("tooLate")

factory.On("Create").Return(something)

testThings(factory)
```

edit: this already got merged: These changes are based on a branch where I did some refactoring, you can see it here https://github.com/vektra/mockery/pull/399. If you don't intend to merge the refactoring PR, I can rebase this branch onto the master.


LandonTClipp:
@grongor  can you please rebase onto master and fix the conflicts? Thanks!

grongor:
@LandonTClipp done :)

pkg/generator.go
```diff
@@ -560,9 +562,39 @@ func (g *Generator) Generate(ctx context.Context) error {
...
 	}
 
+	g.generateConstructor()
+
 	return nil
 }
 
+func (g *Generator) generateConstructor() {
+	const constructor = `
+// %[1]sew%[2]s creates a new instance of %[3]s. It also registers a cleanup function to assert the mocks expectations.
```
LandonTClipp, line 572: Would it be possible you could use golang templates for this? This is kind of difficult to read and I think the templates would make it nicer.

grongor: Well I could do that but to be honest I don't like it :D 
```
diff --git a/pkg/generator.go b/pkg/generator.go
index f26c50c..b969a6a 100644
--- a/pkg/generator.go
+++ b/pkg/generator.go
@@ -8,6 +8,7 @@ import (
 	"go/ast"
 	"go/build"
 	"go/types"
+	"html/template"
 	"io"
 	"os"
 	"path/filepath"
@@ -569,9 +570,9 @@ func (g *Generator) Generate(ctx context.Context) error {
 
 func (g *Generator) generateConstructor() {
 	const constructor = `
-// %[1]sew%[2]s creates a new instance of %[3]s. It also registers a cleanup function to assert the mocks expectations.
-func %[1]sew%[2]s(t testing.TB) *%[3]s {
-	mock := &%[3]s{}
+// {{.FuncFirstLetter}}ew{{.MockNameUpperCase}} creates a new instance of {{.MockName}}. It also registers a cleanup function to assert the mocks expectations.
+func {{.FuncFirstLetter}}ew{{.MockNameUpperCase}}(t testing.TB) *{{.MockName}} {
+	mock := &{{.MockName}}{}
 
@@ -592,7 +593,19 @@ func %[1]sew%[2]s(t testing.TB) *%[3]s {
 		break
 	}
 
-	g.printf(constructor, funcFirstLetter, mockNameUpperCase, mockName)
+	err := template.Must(template.New("").Parse(constructor)).Execute(
+		&g.buf,
+		struct {
+			FuncFirstLetter, MockNameUpperCase, MockName string
+		}{
+			FuncFirstLetter:   funcFirstLetter,
+			MockNameUpperCase: mockNameUpperCase,
+			MockName:          mockName,
+		},
+	)
+	if err != nil {
+		panic("failed to write mock constructor: " + err.Error())
+	}
 }
```

Just now I realized I can simplify the "printf" version, what do you think?
```
diff --git a/pkg/generator.go b/pkg/generator.go
index f26c50c..5deb823 100644
--- a/pkg/generator.go
+++ b/pkg/generator.go
@@ -569,9 +569,9 @@ func (g *Generator) Generate(ctx context.Context) error {
 
 func (g *Generator) generateConstructor() {
 	const constructor = `
-// %[1]sew%[2]s creates a new instance of %[3]s. It also registers a cleanup function to assert the mocks expectations.
-func %[1]sew%[2]s(t testing.TB) *%[3]s {
-	mock := &%[3]s{}
+// %[1]s creates a new instance of %[2]s. It also registers a cleanup function to assert the mocks expectations.
+func %[1]s(t testing.TB) *%[2]s {
+	mock := &%[2]s{}
 
@@ -592,7 +592,7 @@ func %[1]sew%[2]s(t testing.TB) *%[3]s {
 		break
 	}
 
-	g.printf(constructor, funcFirstLetter, mockNameUpperCase, mockName)
+	g.printf(constructor, funcFirstLetter+"ew"+mockNameUpperCase, mockName)
 }

```

grongor: @LandonTClipp Hi, can we please finish this? I'd really love to have the constructors for all my mocks :)

LandonTClipp:
@grongor I tried fixing a merge conflict but messed up a closing bracket, could you fix that? I will merge once the tests pass.

grongor:
@LandonTClipp Done :). Also please note that the Expecter PR did more changes than it should (refactoring, new way to check generated code, ...), specifically `expectedBytes, err := ioutil.ReadFile(getFixturePath("mocks", "expecter.go"))`. I'm not saying that it is bad...just that this is the first place in the codebase that it is done this way, so I had to look around for some time until I managed to find a way to fix the errors after rebasing. I also reverted part of the Expecter PR because it caused panics during tests (it's in a separate commit, please check).

LandonTClipp:
Thanks @grongor, can you elaborate on what you mean by "new way to check generated code"?

pkg/generator.go
```diff
@@ -699,6 +703,32 @@ func (_c *{{.CallStruct}}) Return({{range .Returns.Params}}{{.}},{{end}}) *{{.Ca
...
+	return mock
+}
+`
+
+	funcFirstLetter := "N"
+	mockName := g.mockName()
+
+	for _, firstLetter := range mockName {
+		if unicode.IsLower(firstLetter) {
+			funcFirstLetter = "n"
```
LandonTClipp, line 723: I've seen a repeated pattern of where we manually modify the first letter depending on whether it's exported or not. Might make sense have a helper function that will make a string exported/non-exported depending on the state of some other variable/function name.

Makefile
```diff
@@ -1,11 +1,7 @@
 SHELL=bash
 
 .PHONY: all
-all: clean fmt mocks test install docker integration
-
-.PHONY: clean
-clean:
-	rm -rf mocks
+all: fmt mocks test install docker integration
```
LandonTClipp, line 4: 👍🏻

pkg/generator_test.go
```diff
@@ -80,10 +80,11 @@ func (s *GeneratorSuite) checkGenerationWithConfig(
...
 	actualLines := strings.Split(string(actual), "\n")
 
-	// Error out at first unmatched line
-	for i := range actualLines {
-		s.Equal(expectedLines[i], actualLines[i])
-	}
+	s.Equal(
+		expectedLines, actualLines,
+		"The generator produced unexpected output.",
+	)
```
LandonTClipp, line 86: 👍🏻 Thanks for catching this.

LandonTClipp (approved):
PR looks good, I only have one comment about the export/unexported name logic but that doesn't hold up this PR. If you want extra credit you can make the helper function :)

LandonTClipp:
Because we're dealing with some issues regarding go1.18 builds I'm going to pause this PR to minimize changes, but once that settles down this is good to go.

grongor:
> Can you elaborate on what you mean by "new way to check generated code"?

this file: `pkg/fixtures/mocks/expecter.go`

The rest of the generated mock files live in `mocks` directory. The generator test contains expected code directly in the test file `pkg/generator_test.go`, and there is only one test using an "external" file `pkg/fixtures/mocks/expecter.go`. Maybe there is a good reason for it - I don't know. I just wanted to point out that it kind of does things differently than the rest of the tests.

> Because we're dealing with some issues...

No problem, I'm glad that it's finally moving :) Thanks!

pkg/generator.go
grongor: I didn't but yeah if you know about such places it would make sense to make a helper function :) I know about the `upperFirstOnly`, but it always changes the string (doesn't check another).

LandonTClipp: If you could create a helper function in a new pkg/strings.go or whatever name makes sense, that would be great. I'm sure we'll use it again. Thanks!

LandonTClipp:
@grongor merged and deployed. One additional matter, could you document this in the README as a new feature that is available? I'm sure others will benefit from it.

grongor:
@LandonTClipp thank you! I'll look into updating the README soon, as well as the helper function you mentioned (https://github.com/vektra/mockery/pull/406#discussion_r840819968).

grongor:
https://github.com/vektra/mockery/pull/450

pkg/generator.go
grongor: See this: https://github.com/vektra/mockery/pull/452

PR #418, opened 2021-11-07
feature: parse multiple interfaces with --name

positiveblue:
Allow the `--name` param to specify multiple interfaces/regex separated
with `,`.

### DIFF
--- a/cmd/mockery.go
+++ b/cmd/mockery.go
@@ -131,8 +131,6 @@ func initConfig() {
 	}
 }
 
-const regexMetadataChars = "\\.+*?()|[]{}^$"
-
 type RootApp struct {
 	config.Config
 }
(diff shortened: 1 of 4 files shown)


cmd/mockery.go
```diff
@@ -145,9 +143,33 @@ func GetRootAppFromViper(v *viper.Viper) (*RootApp, error) {
 	return r, nil
 }
 
+func getFilters(names []string, log zerolog.Logger) ([]*regexp.Regexp, bool) {
```
LandonTClipp, line 146: The log should be passed through `ctx`, see some of the other examples in the code of how we do that.

cmd/mockery.go
```diff
@@ -145,9 +143,33 @@ func GetRootAppFromViper(v *viper.Viper) (*RootApp, error) {
...
+
+	var hasRegex bool
+	var filters []*regexp.Regexp
+
+	for _, name := range names {
+		if strings.ContainsAny(name, regexMetadataChars) {
+			hasRegex = true
+			filter, err := regexp.Compile(name)
+			if err != nil {
+				log.Fatal().Err(err).Msgf("Invalid regular expression provided to -name")
```
LandonTClipp, line 158: The error should be returned and bubbled up the stack. I generally avoid `log.Fatal()` as it's non-graceful.

cmd/mockery.go
```diff
@@ -176,19 +198,18 @@ func (r *RootApp) Run() error {
...
 		recursive = r.Config.Recursive
-		if strings.ContainsAny(r.Config.Name, regexMetadataChars) {
-			if filter, err = regexp.Compile(r.Config.Name); err != nil {
-				log.Fatal().Err(err).Msgf("Invalid regular expression provided to -name")
-			} else if r.Config.FileName != "" || r.Config.StructName != "" {
-				log.Fatal().Msgf("Cannot specify --filename or --structname with regex in --name")
-			}
-		} else {
-			filter = regexp.MustCompile(fmt.Sprintf("^%s$", r.Config.Name))
+		names := strings.Split(r.Config.Name, ",")
```
LandonTClipp, line 201: I think we can instead use https://pkg.go.dev/github.com/spf13/pflag#StringSlice to make the `--name` flag an explicit multi-value flag.

LandonTClipp (changes requested):

LandonTClipp:
Closing this for inactivity but please re-open if you manage to fix the points I mentioned.

PR #422, opened 2021-11-17
Fix: avoid package name collision with inPackage (#291)

### LINKED ISSUE #291: Invalid mock generated with the same package names

i-sevostyanov:
Fixes #291.

Tired of waiting, so I copied everything from https://github.com/vektra/mockery/pull/402, sorry.


fterrag:
Can we get this PR reviewed, please? Thanks!

peppelan:
same here, would be great, thankyousomuchplease

LandonTClipp:
@strideynet sorry for the delay, life gets in the way sometimes, I'm trying to work through a backlog of stuff. Can you write in the description more succinctly what the solution to the original problem is? This is just to aid me so I can know better what I'm looking at.

LandonTClipp:
@i-sevostyanov  @strideynet sorry for the delay, life gets in the way sometimes, I'm trying to work through a backlog of stuff. Can you write in the description more succinctly what the solution to the original problem is? This is just to aid me so I can know better what I'm looking at.

i-sevostyanov:
I think this issue describes the problem in detail: [Invalid mock generated with the same package names](https://github.com/vektra/mockery/issues/291) 🤓

serivkos:
@LandonTClipp I would appreciate if you could look over this request and merge it with master. I’ll be waiting for that. Thanks

fterrag:
We're currently working around this with some `sed` magic - looking forward to a fix!

pkg/generator.go
```diff
@@ -104,7 +104,10 @@ func (g *Generator) addPackageImportWithName(ctx context.Context, path, name str
 
 func (g *Generator) getNonConflictingName(path, name string) string {
 	if !g.importNameExists(name) {
-		return name
+		// do not allow imports with the same name as the package when inPackage
+		if !g.InPackage || g.iface.Pkg.Name() != name {
+			return name
```
LandonTClipp, line 109: Let's collapse this if statement into one

```go
if !g.importNameExists(name) && (!g.InPackage || g.iface.Pkg.Name() != name)
```

pkg/generator.go
```diff
@@ -121,7 +124,10 @@ func (g *Generator) getNonConflictingName(path, name string) string {
 	for i := 1; i <= numDirectories; i++ {
 		prospectiveName = strings.Join(cleanedDirectories[numDirectories-i:], "")
 		if !g.importNameExists(prospectiveName) {
-			return prospectiveName
+			// do not allow imports with the same name as the package when inPackage
```
LandonTClipp, line 127: Same, let's collapse into one if statement.

pkg/fixtures/example_project/foo/collision.go
```diff
@@ -0,0 +1,7 @@
+package foo
+
+import "github.com/vektra/mockery/v2/pkg/fixtures/example_project/bar/foo"
+
+type Collision interface {
```
LandonTClipp, line 5: Maybe add a comment in this describing what the intended situation here is (obviously, package and import having the same name).

LandonTClipp (changes requested):
Thanks for the PR, just a few minor comments and we can get this merged.

pkg/generator.go
i-sevostyanov: Done

pkg/generator.go
i-sevostyanov: Done

pkg/fixtures/example_project/foo/collision.go
i-sevostyanov: I renamed the interface, so now it has a more explicit name

LandonTClipp: Thanks @i-sevostyanov , tests seem to be failing unfortunately. I'll let you take care of that.

i-sevostyanov: Oh, sorry for that. Fixed 🤓

LandonTClipp (approved):
Looks good!

LandonTClipp:
@i-sevostyanov sorry one more thing :) the coverage report is showing some missing coverage. 

https://app.codecov.io/gh/vektra/mockery/compare/422/changes

Could you add tests for the cases that are missing? PRs generally should be increasing coverage, as a matter of principle.

i-sevostyanov:
Done 🤓

Issue #416, opened 2021-11-04
internal error when running mockery against Go tip


palash25:
so our CI jobs on Golang tip `go version devel go1.18-4d7bf41 Tue Nov 2 13:43:24 2021 +0000 linux/amd64`  started failing recently with this error 

```
 go generate ./...
2021/11/02 14:26:16 internal error: package "context" without types was imported from "github.com/orgname/reponame/pkg"
pkg/deps.go:16: running "mockery": exit status 1
```

Below is a snippet from the file where mockery is called (can't paste the complete code because the repo is private)
```go
import (
"context"
)

//go:generate mockery -name=identityProviderService -case=snake -inpkg -testonly
type identityProviderService interface {
	FindUser(ctx context.Context, login string) (*User, error)
}
```

any idea whats going on here?

siraphobk:
I had this problem too. However, I forked the repo and update dependencies in go.mod and it's working now. #425 should fix it.

SubhramRanaRZP:
I am using go1.19 , and while using `mockery` tool to generate mocks, I am facing the same. @siraphobk can you help me . thanks

palash25:
@SubhramRanaRZP what is your mockery version? I remember the issue was resolved for us after we upgraded mockery

tarasmadan:
One more example:

go version go1.20-pre3

```
$ mockery --dir vm/proxyapp/ --name subProcessCmd --exported
18 Oct 22 08:51 UTC INF Starting mockery dry-run=false version=v2.14.0
18 Oct 22 08:51 UTC INF Walking dry-run=false version=v2.14.0
2022/10/18 08:51:27 internal error: package "context" without types was imported from "github.com/google/syzkaller/vm/proxyapp"
```

Target interface is [here](https://github.com/google/syzkaller/blob/master/vm/proxyapp/init.go).

sicoyle:
Hi! I just wanted to comment on here that using `v2.10.0` with `go1.19.2` also gave me similar errors as commented on in this issue. See below:

```bash
mockery --all
09 Nov 22 15:44 CST INF Starting mockery dry-run=false version=v2.10.0
09 Nov 22 15:44 CST INF Walking dry-run=false version=v2.10.0
2022/11/09 15:44:34 internal error: package "fmt" without types was imported from "<I removed some stuff here>"
```

The fix was to upgrade to the latest `mockery` version for me. I am able to successfully run `mockery --all` using `mockery` `v0.14.1`.

[...]

PR #425, opened 2021-12-16
Upgrade golang.org/x/tools to v0.1.8

fterrag:
Fixes https://github.com/vektra/mockery/issues/416 by upgrading golang.org/x/tools to v0.1.8

@palash24


fterrag:
@LandonTClipp

polothy:
For me, this PR did not work _until_ I built mockery locally with `go1.18beta1 build .` - thanks for getting me unblocked.

I'm kinda just bumbling around on the weekend here, so take that into consideration when reading my comment 😄

PR #438, opened 2022-03-28
Add option to let mock accept no arguments of return value provider function

comphilip:
Given this is in `proxy.go`

```go
package test

import "context"

type Proxy interface {
	passthrough(ctx context.Context, s string) string
}

```

Run: `mockery --return-func-no-arg` and the following will be output to `mocks/Proxy.go`:

```go
// Code generated by mockery . DO NOT EDIT.

package mocks

import (
	context "context"

)

// Proxy is an autogenerated mock type for the Proxy type
type Proxy struct {
	mock.Mock
}

// passthrough provides a mock function with given fields: ctx, s
func (_m *Proxy) passthrough(ctx context.Context, s string) string {
	ret := _m.Called(ctx, s)

	var r0 string
	if rf, ok := ret.Get(0).(func(context.Context, string) string); ok {
		r0 = rf(ctx, s)
	} else if rf, ok := ret.Get(0).(func() string); ok {
		r0 = rf()
	} else {
		r0 = ret.Get(0).(string)
	}

	return r0
}
```

No arguments return provider function also acceptable. **No extra typing needed if return value does not depends on arguments.**

```go
import . "github.com/stretchr/testify/mock"

// with arguments
Mock.On("passthrough", mock.AnythingOfType("context.Context"), mock.AnythingOfType("string")).Return(func(ctx context.Context, s string) string {
    return s
})

// without arguments
Mock.On("passthrough", mock.AnythingOfType("context.Context"), mock.AnythingOfType("string")).Return(func() string {
    return "test"
})
```


Gevrai:
If return value does not depend on arguments you can already pass the value directly instead of using a match function.
```
Mock.On("passthrough", ... ).Return("test")
```
Or is there another issue with this that I'm missing?

comphilip:
@Gevrai agree with you. I search in my code base, and there is only 4 `.Return(func(` and only one uses no arguments function provider.

PR #441, opened 2022-03-31
fix: allow configuring flags with "-" as environment variables

cfstras:
Dashes are not allowed in variable names in most shells. While you can work around this with `env`, this doesn't work when you want to `export` it as a global env var (because shells will also ignore env variables with dashes in them).

For example MOCKERY_DISABLE_VERSION_STRING.

So, this workaround works:
```
➜ env "MOCKERY_DISABLE-VERSION-STRING=true" mockery showconfig | grep versionstring
disableversionstring: true
```

But this doesn't:
```shell
➜ MOCKERY_DISABLE_VERSION_STRING=true mockery showconfig | grep versionstring
disableversionstring: false
```

After this fix, we can use it:
```shell
➜ MOCKERY_DISABLE_VERSION_STRING=true ./mockery showconfig | grep versionstring
disableversionstring: true
```

### DIFF
--- a/cmd/mockery.go
+++ b/cmd/mockery.go
@@ -106,6 +106,7 @@ func Execute() {
 
 func initConfig() {
 	viper.SetEnvPrefix("mockery")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	viper.AutomaticEnv()
 
 	if cfgFile != "" {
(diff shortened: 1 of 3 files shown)


LandonTClipp:
Thanks for the find @cfstras , could you add a unit test to this? Once that's done we can merge.

cfstras:
@LandonTClipp added one that covers all the config vars, and found an unused field in the process :)

LandonTClipp:
Thanks for the contribution @cfstras, getting it built now.

PR #450, opened 2022-04-19
Update readme (new constructors, mention expecter, small improvements)

grongor:


README.md
```diff
@@ -287,6 +320,27 @@ requesterMock.EXPECT().
...
+Mockery v2.11 introduces constructors for all mocks. This makes instantiation and mock registration a bit easier and
+less error-prone (you won't have to worry about forgetting the `AssertExpectations` method call anymore).
+
+Before v2.11:
+```go
+factory := &mocks.Factory{}
+defer factory.AssertExpectations(t)
+```
+
+You can use the type-safe expecter interface as such:
```
LandonTClipp, line 337: This might be a terminology overload, the type-safe expecter methods are opt-in and disabled by default, so this isn't strictly true.

README.md
```diff
@@ -287,6 +320,27 @@ requesterMock.EXPECT().
 	Call.Return(func(path string) string { return "result for " + path }, nil)
 ```
 
+Mock constructors
```
LandonTClipp, line 323: You'll also want to add this to the table of contents in the appropriate area.

README.md
grongor: LOL this is totally a mistake - I copied the section from the expecter and then modified it...so I guess I completely missed this one line :D 

And by the way, I would try to push the users to the expecter interface as much as possible, to be honest :) It's far superior way to work with mocks IMO. Just a side note... :)

LandonTClipp (approved):
Looks good, thanks!

PR #454, opened 2022-04-20
Move generated ExpecterTest to mocks directory

Gevrai:
Description
-------------

Fixes the location of the `--expecter` tests mocks, as requested in https://github.com/vektra/mockery/pull/396#issuecomment-1101518339

## Type of change

Test files moved to correct location, new behaviour change for users.

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [x] 1.16
- [ ] 1.17
- [ ] 1.18

How Has This Been Tested?
---------------------------

Tested with `go test ./...` on full project

Checklist
-----------


Gevrai:
@LandonTClipp As far as I can see the command `make mocks` doesn't generate all the right ones in the mocks file, is that expected? I didn't add this one for the expecter because it seemed outdated, but I might be wrong.

mocks/pkg/fixtures/ExpecterTest.go
```diff
@@ -1,18 +1,26 @@
-// Code generated by mockery. DO NOT EDIT.
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
grongor, line 1: This looks like you used the wrong version to generate the mocks...strange.

grongor:
@Gevrai I think the mocks are generated correctly (but I might be wrong, please post details if you are sure). The issue here is that you combine the generated mocks and the fixtures in an unusual way for this repo IMO. All generator tests have their expected output in the generator test itself. I think the best way forward is to just update the expecter test to behave the same as other tests, and I'll look into it later and try to move these fixed strings to separate files.

The problem with what you did (I think) is that you say that you expect the generated code to be equal to the mock file, but the mock file itself is generated by the generator, so this test doesn't really validate anything. I might be wrong though...

mocks/pkg/fixtures/ExpecterTest.go
LandonTClipp: Agreed, could you verify what mockery you're using to build the mocks? These version strings should be disabled so this is telling me that your mockery probably wasn't reading the config properly. (maybe we should re-introduce the version numbers?)

pkg/generator_test.go
```diff
@@ -240,7 +240,7 @@ func NewRequester(t testing.TB) *Requester {
 }
 
 func (s *GeneratorSuite) TestGeneratorExpecterComplete() {
-	expectedBytes, err := ioutil.ReadFile(getFixturePath("mocks", "expecter.go"))
+	expectedBytes, err := ioutil.ReadFile(getMocksPath("ExpecterTest.go"))
```
LandonTClipp, line 243: The thing I agree with @Gevrai here is that it's odd to be asserting the output of mockery with the output of mockery (you see what I mean?). You are generating `ExepcterTest.go` in the `mocks` package, then you generate it again in this test and assert that it's the same. It's not _necessarily_ wrong to do it this way, but none of the other tests do it this way (see `TestGeneratorFunction` below). 

Although, I can see an argument to be made that because the `Expecter` interface is so verbose in the generated code, that cluttering the tests with expected code would be too much. But I think that for this case, we really should have the expected code as a raw string in the test.

LandonTClipp (changes requested):

pkg/generator_test.go
grongor: I think we should stick with how the tests are done for now (strings directly in the test), but definitely look for improvements later on. As you mentioned, asserting mockery with mockeries output isn't correct, but we definitely can improve the readability of the tests, maybe even the way @Gevrai described. But it definitely is out of the scope of this, and it's a bigger decision that should be done correctly (so that we don't face it again in a year, hopefully :D ).

LandonTClipp: I do really like the idea of keeping unit tests for the generated mocks (`mocks/pkg/fixtures/ExpecterTest_test.go`), that can probably stay. I think we should have more of these unit tests on the generated mocks as another layer.

LandonTClipp: @Gevrai let me know if you could make the change (expected output as a string literal in the test). We can get it merged once that's settled. Thanks!

Gevrai: Hey sorry! Things got in the way a bit, I'll have it done this evening.

Just to make sure
- Keep test in mocks/pkg/fixtures/ExpecterTest_test.go
- Inline generated expecter check instead of reading from file

LandonTClipp: No worries thanks again for the contributions. Yeah those two points should be it!

Gevrai: Done, should be good now.

FYI I've changed the name of the interface (ExpecterTest -> Expecter) for coherence with others, and added the expecter to be generated by Makefile so that the unit tests have something to run on!

mocks/pkg/fixtures/ExpecterTest.go
Gevrai: I hadn't used the Makefile, fixed now

Gevrai:
branch rebased

LandonTClipp:
Thanks @Gevrai you read my mind.

LandonTClipp (approved):

Issue #442, opened 2022-03-31
Release v2.10.1 hangs on Starting mockery dry-run=false version=v2.10.1


DillonStreator:
Noted when running `mockery --all --inpackage --case snake`

DillonStreator:
resolved for now by pinning previous version
`go install github.com/vektra/mockery/v2@v2.10.0`

LandonTClipp:
I think I know where this might be coming from. My suspicion isn't that it's hanging, but that it's taking a lot longer than normal. Probably from this: https://github.com/vektra/mockery/commit/4e181be712c853258e64b29a9797f5f596661449

Please try v2.10.2, if that does work I will revert the above commit.

Tochemey:
Yeah I have observed it takes longer than before to generate the mocks.
<img width="476" alt="image" src="https://user-images.githubusercontent.com/1633645/161351967-f56c97ce-f182-47c8-b022-ed8d1c6e0554.png">

Tochemey:
At times it does hang even the lastest release.

taylow:
I am having the same issue on `darwin/arm64` (Apple M1) running `mockery --all --keeptree` and have noticed the memory usage just keeps growing.

<img width="269" alt="image" src="https://user-images.githubusercontent.com/19749421/161380297-10e608f9-144a-4174-82e7-3d19e7cddd99.png">

ludusrusso-rc:
Same here

LandonTClipp:
Folks can you try the latest release, I have reverted a change that I believed was causing the runtime issue. @emmanuel099 is showing a list of versions that work: https://github.com/vektra/mockery/pull/444

taylow:
Just installed 2.10.4 and have a new issue, is there a known fix for this error?

<img width="533" alt="image" src="https://user-images.githubusercontent.com/19749421/162399833-ad00158b-14c4-47b2-a841-9313efc3ae6d.png">

LandonTClipp:
@taylow could you provide a reproducer for this issue?

taylow:
@LandonTClipp happy to provide as much info as I can!

My current setup is:
- `go version go1.18 darwin/arm64`
- Mockery version `v2.10.4` installed via Brew
- I'm using an M1 Mac `darwin/arm64`
- I'm running the command `make mocks` via a Makefile, which calls `mockery --all --keeptree`
- I currently have Mocks that are outdated which need regenerating

If there is anything else you need/could help here please let me know!

[...]

PR #456, opened 2022-04-20
Add support for generating mocks for generic interfaces

### LINKED ISSUE #451: mockery fails to generate mocks for interfaces that use generics
Description
-------------

This is a meta ticket to track issues with 1.18 generics.

https://github.com/vektra/mockery/issues/442
https://github.com/vektra/mockery/issues/439

Version
--------

2.11.0

cruickshankpg:
Description
-------------

This adds support for generating mocks for generic interfaces. It does bump the go version to 1.18 so not sure how you want to handle this. 

- Fixes #451

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18

How Has This Been Tested?
---------------------------

I've added unit tests for different types of generic interfaces, happy to add any others you think are worthwhile.

Checklist
-----------


LandonTClipp:
Thanks for the PR, this is looking great!

LandonTClipp:
@cruickshankpg could you fix the merge conflicts?

pkg/generator.go
```diff
@@ -88,6 +102,13 @@ func (g *Generator) addImportsFromTuple(ctx context.Context, list *types.Tuple)
 	}
 }
 
+func (g *Generator) addPackageScopedType(ctx context.Context, o *types.TypeName) string {
```
LandonTClipp, line 105: Let's give the function a description, something like

```
// addPackageScopedType returns the appropriate string representation for the object TypeName. The string may either be the unqualified name (in the case the mock will live in the same package as the interface being mocked) or the fully qualified pathname (in the case the type lives in a package external to the mock).
```

pkg/generator.go
```diff
@@ -225,6 +246,36 @@ func (g *Generator) mockName() string {
...
+		case *types.Named:
+			qualifiedParams = append(qualifiedParams, fmt.Sprintf("%s %s", param.String(), g.addPackageScopedType(ctx, constraint.Obj())))
+		case *types.Interface:
+			qualifiedParams = append(qualifiedParams, fmt.Sprintf("%s %s", param.String(), constraint.String()))
+		}
+	}
+	return fmt.Sprintf("[%s]", strings.Join(qualifiedParams, ", "))
+}
+
+func (g *Generator) typeParams() string {
```
LandonTClipp, line 267: `s/typeParams/getInstantiatedTypeString/g`

Comment like:

```
// getTypeParams returns the "instantiated" type names for a given constraint list. For instance, 
// if your interface has the constraints `[S Stringer, I int, C Comparable]`, this method would return:
// `[S, I, C]`

File: pkg/generator.go [hunk 3]
```diff
@@ -329,11 +380,12 @@ type namer interface {
...
-		o := t.Obj()
-		if o.Pkg() == nil || o.Pkg().Name() == "main" || (!g.KeepTree && g.InPackage && o.Pkg() == g.iface.Pkg) {
-			return o.Name()
+		return g.addPackageScopedType(ctx, t.Obj())
+	case *types.TypeParam:
+		if t.Constraint() != nil {
+			return t.Obj().Name()
 		}
-		return g.addPackageImport(ctx, o.Pkg()) + "." + o.Name()
+		return g.addPackageScopedType(ctx, t.Obj())
```
LandonTClipp, line 388: Can you explain the difference between the `t.Constraint() != nil` block and the other block?

LandonTClipp (changes requested):
I have a few comments on this. I have a few other thoughts that I will share in the coming days. Thanks again.

LandonTClipp:
Thanks for the update @cruickshankpg.

So I have thought about this for a while and I want to add a new release before this with an explicit deprecation notice in the logs for `go install`. I've added notes in the readme that this will break but I want to leave a deprecation notice out for possibly a month just to make it as clear as possible.

Sorry to delay this but I think this is the right way forward.

slessard:
If not `go install` then how should one install this new version of Mockery? (Even an unsupported suggestion would be welcome.)

LandonTClipp:
The only recommended/supported methods are:

1. Downloading the pre-built binaries
2. Using the docker image
3. Using homebrew

You _can_ use go install/go get if you want, but there are two caveats here:

1. We will not support issues with this method going forward
2. You will have to keep your local Golang binary at a version greater than or equal to what's listed in our `go.mod`. As long as your go version is up to date, it will _probably_ work most of the time. But please don't use it for production stuff as any time we add support for new language features, it will likely break your workflows.

fastcat:
What about `go run github.com/vektra/mockery/v2 ...` esp. from within a `//go:generate` directive? That will always use the same version you have in your `go.mod`. It wastes a bit of linker time, but not much. It also avoids problems with different modules requiring different versions of this package's CLI to be installed somewhere "global".

LandonTClipp:
@fastcat using any golang command that compiles the source code using a local go binary isn't supported. You are certainly free to use the command if you find it useful for your purposes but I strongly recommend against it.

LandonTClipp:
@fastcat

> That will always use the same version you have in your go.mod

That always uses the same golang version specified in go.mod? Does golang download a more recent binary if your local is too old?

guenhter:
@cruickshankpg go 1.18.2 was released yesterday. Maybe this solves https://github.com/golang/go/issues/51629 for you and you can get rid of pinning the version. Worth a try I'd say.

fastcat:
@LandonTClipp ok, thanks for clarifying what the risk/concern is. It wasn't clear before, and seemed like a `go install` or `go run` created binary might malfunction, which I have seen be a problem for other tools that have similar disclaimers! Having an open source tool that says "we don't support _you_ compiling this tool from source, only _us_ doing it" without further context is _weird_.

Pinning the version in my `go.mod` seems like it will avoid all these issues in my case -- there won't be any "surprises" unless I change it. And while I'm sure many others will disagree, my 2 cents is that I don't see a problem with "you need to update to a Go version released many months ago" as a problematic thing for a feature bump in a tool like this :)

LandonTClipp:
I'll go ahead and merge this in the coming days and give the version a `-beta.0` suffix. I'll release a non-beta tag after June 1st. @cruickshankpg could you rebase on master to confirm the tests work?

cruickshankpg:
> @cruickshankpg go 1.18.2 was released yesterday. Maybe this solves [golang/go#51629](https://github.com/golang/go/issues/51629) for you and you can get rid of pinning the version. Worth a try I'd say.

yep I'm not seeing it after upgrading to 1.18.2. I've removed the upgrade to an interim x/tools version

pauleibye:
FWIW on go version 1.18.2 I still get `2022/05/13 09:40:16 Unexpected package creation during export data loading` after pulling this pr and running `mockery --all -r --case snake`

ashyxz:
> > @cruickshankpg go 1.18.2 was released yesterday. Maybe this solves [golang/go#51629](https://github.com/golang/go/issues/51629) for you and you can get rid of pinning the version. Worth a try I'd say.
> 
> yep I'm not seeing it after upgrading to 1.18.2. I've removed the upgrade to an interim x/tools version

For anyone coming to this, I had to `replace golang.org/x/tools => github.com/reillywatson/tools v0.1.10-0.20220322163840-50e5cab0be89` per [this](https://github.com/vektra/mockery/issues/442#issuecomment-1097040090) comment

cruickshankpg:
Retested and I was still seeing `Unexpected package creation during export data loading`  so have reinstated the pin to an interim x/tools version that contains the fix

slessard:
> The only recommended/supported methods are:
> 
> 1. Downloading the pre-built binaries

Is there an easy link where I can always download the latest release similar to how `go install github.com/vektra/mockery/v2@latest` always gets the latest release?

PH9:
I using GitLab CI with go image how use mockery on CI then?

When go image has no docker inside, and go image has no brew to install.

mfrister:
@LandonTClipp

> The only recommended/supported methods are:
> 
> 1. Downloading the pre-built binaries
> 2. Using the docker image
> 3. Using homebrew

Unfortunately, no longer supporting `go install` means projects using mockery can no longer use tool versioning via a `tools.go` and `go.mod`, as proposed in the [Go Wiki](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module).

Versioning via go.mod has the benefit of automated dependency update tools like [Renovate](https://github.com/renovatebot/renovate) being able to automatically create MRs for tool updates.

Not being able to use `go install` then means having to make a decision between manually updating mockery (when installing pre-built binaries) and no longer using `go generate ./...` in CI jobs checking that all generated code is up-to-date (when using a separate CI job with the mockery Docker image that doesn't have the other tools installed).

I don't have a solution ready, but maybe this is an argument for looking for a better solution. I'd be ok with requiring an up-to-date Go version for the newest mockery versions, I'm just afraid that generally declaring `go install` as not supported may result in decisions that may break `go install` for other reasons than using an older Go versions in the future.

LandonTClipp:
@slessard we can probably modify the asset names so that you can reference the `latest` URL. For instance currently you'd use a URL like:

https://github.com/vektra/mockery/releases/download/v2.12.3/mockery_2.12.3_Darwin_x86_64.tar.gz

But since the tar file has the version name it makes it hard to download the latest as the name will always change. So instead we could do something like:

https://github.com/vektra/mockery/releases/latest/mockery_Darwin_x86_64.tar.gz

How does that sound?

slessard:
> @slessard we can probably modify the asset names so that you can reference the `latest` URL. For instance currently you'd use a URL like:
> 
> https://github.com/vektra/mockery/releases/download/v2.12.3/mockery_2.12.3_Darwin_x86_64.tar.gz
> 
> But since the tar file has the version name it makes it hard to download the latest as the name will always change. So instead we could do something like:
> 
> https://github.com/vektra/mockery/releases/latest/mockery_Darwin_x86_64.tar.gz
> 
> How does that sound?

That is certainly a positive change. Also an apt-get package for certain linux distributions would be nice

wspowell:
> This change could/will break anyone who is using go install to download mockery, or any installation method that requires compilation using a locally-installed go binary. 

This is proven false with your follow up response.
> Intrinsically, the issue is that we break semantic versioning if minor bumps might require people to upgrade their golang version. 

So it _will_ work as long as the local golang binary matches at least the go.mod version.

For a project whose _sole_ purpose is to closely follow a language spec and generate code based on that language's changes, I believe using semantic versioning as a reason for abandoning a core language design (`go install`) is extremely misguided. Semantic versioning is at the core of the argument to abandon `go install` support entirely because anyone bumping their minor version might have their workflows broken. And yet, while using this argument, are willing to break that contract for the go1.18 update.

Programming gets messy sometimes (a lot of times). For a project like this, I think it is acceptable to bump minor versions for new golang versions that contain minor breaking changes when the project is forced to in order to keep on producing correct mocks (otherwise, no go.mod change is necessary). When a major change like generics comes along, if there is any time to bump the major version this is it. While my statement is my own opinion, I believe Golang devs will understand and be willing to work with you on these aspects regarding semantic versioning since by definition you are not in complete control of how you can semantically version your project.

That being said, I urge you to remove the "official deprecation" label of `go install` on the project. A caveat can be added that states that a golang binary update might be required for minor revisions, if that helps. For those of us that rely on your project for work, it either forces us to abandon `go install` completely due to potentially instability (even though it sounds like it will work just fine) or potentially persuade devs to find alternatives, which would be difficult, or avoid the project in the first place. Either way, it shifts that burden onto the consumers of your project, which I find unacceptable inside golang's ecosystem.

LandonTClipp:
Thanks for your insight @wspowell .

> This is proven false with your follow up response.

I'm not sure I follow. It will break anyone who doesn't have an up-to-date Golang version installed. Yes it is corrected by updating your Golang version but intrinsically if we are requiring users to modify their environment to support updates to this project outside of major version bumps, we have broken semver.

> I believe using semantic versioning as a reason for abandoning a core language design (go install) is extremely misguided.

The stance I am taking is that officially we do not want to tell people to use `go install` for the reasons I've mentioned. I'm okay with placing a caveat that we will do our best to ensure `go install` works, but that users need to understand the implications for their workflows. Golang users are fairly capable folk so I'm not worried about that, I just don't want people to become upset that we broke their workflows because they are having to upgrade golang because of a feature they might not even need.

> And yet, while using this argument, are willing to break that contract for the go1.18 update.

We have already publicized that `go install` is not tracked in the semver for quite a long time in the readme. This is simply been the first time we've been forced to enforce it. The contract has already been "broken" for months/years.

> or avoid the project in the first place

Certainly I want to avoid pushing you away :) 

> Either way, it shifts that burden onto the consumers of your project, which I find unacceptable inside golang's ecosystem.

There is a burden placed on the users either way. I'm simply picking a burden that I feel is in the best interest of semantic and logical correctness. 

I do hear your concerns and I greatly appreciate your input. I'm amenable to the idea of perhaps not deprecating it entirely, but recommending people not to use it. And if folks want to use it, then I can point to our stance when they submit tickets asking why their workflows are broken.

fastcat:
To me there are _huge_ differences between:

1. `go install` may produce a binary that malfunctions
2. `go install` may produce a binary that generates source code that fails to compile with your version of `go`
3. `go install` may fail to compile a binary

Items (2) and (3) create _very_ clear errors for the user. Something fails very clearly, and you either upgrade `go` or you downgrade `mockery`. And from what you've said, my inference is that (3) is the sole concern here, yes?

While I disagree with your stance on (not) requiring folks to update Go, I recognize it's something on which reasonable people may disagree :slightly_smiling_face:. What I _do_ ask is that the deprecation notice be made _clear_ about the risks and reasons for why `go install` is discouraged. The existing notice and linked PR comment does _not_ make clear what the risk is.

Absent that clarity, the current documentation reads as equivalent to "despite this being an _open source_ project, we do not support users compiling it for themselves". That would be a very anti-open-source thing to say, and from the more detailed contents of this thread (as opposed to the one linked comment from the README), it's obviously _not what you mean_.

So, in the interest of staying positive & being helpful, some suggested wording that I think would clarify this:

> If you choose to compile mockery from source, such as with `go install`, you may need to upgrade your environment to have at least the version of Go declared in this project's `go.mod`, even if you are creating mocks for an older version of Go. If your version of Go is at least that specified in this project's `go.mod`, then compiling mockery from source should produce a working binary, and the mocks it generates should be compatible with the version of Go used in your project, i.e. if the interfaces you are mocking work with a given version of Go, then the generated mocks should be compatible with that same version.

irridia:
I tend to agree with @wspowell. The following is IMHO.

It feels like all of the issues discussed in this PR thread can be distilled down to the consequences of one decision: not wanting to use semantic versioning as intended because "it's a bit much". If you're releasing software that is no longer backwards-compatible, you increase the major version. This is widely understood, even assumed, and I would posit a critical foundation of modern software engineering. I agree bumping a major version just for this support is _a lot_, but it's the only answer that fixes more than it breaks.

While major version tagging is a bit hairy in go, I think it's safe to assume consumers (developers) now understand the  tagging/versioning semantics of the go ecosystem more reliably than they will know the one-off decisions of arbitrary GitHub projects that could be direct or _indirect_ dependencies on their work. And they would likely prefer to understand one integrated versioning and tooling mechanism than several potential mechanisms (binary, docker, etc) each with their own mockery-specific version choice caveats.

Removing guarantees of support for a foundational part of go tooling is a slippery slope (looking at you, `golangci-lint`), and the hesitance to call a spade a spade doesn't help.

I happen to use `brew` for this, but if I had to use this tool in automation I would not feel comfortable with "downloading a binary at a pinned hash" or "setting up a new docker compose flow to integrate a pinned docker image". Using hardcoded versions isn't a responsible option to me—it again (see: slippery slope) defeats the purpose of semantic versioning. I'll also hand-wave at the security implications of this approach as an exercise left to the reader. ;)

nashikb:
Thanks for the points and followup!

Issue #395, opened 2021-07-08
Exclude specific subfolder from --all


lordzsolt:
Mostly a followup to #255, with the introduction of go modules, dependencies are in the `vendor` subfolder.

With the command `mockery --all`, mocks will be generated for all interfaces of third party dependencies.

It would be great to have some way of excluding this folders (and potentially other ones). 

Is there a way to do this already?

vasconcelosvcd:
+1

Trim21:
+1

AntonioYuen:
+1

alec-w:
I'd also find this feature useful.

Taking a look at the code it looks like the behaviour change would be within https://github.com/vektra/mockery/blob/8174e46cfded8b352b05339478e6263e30cbecc8/pkg/walker.go#L65

I'd have a stab at a PR on this, but an open question is - should the exclude be just for top level directories or any subdirectory matching a given pattern (and if the latter, how to make sure `root1/foo` is included but `root2/foo` is excluded?)?

LandonTClipp:
Thanks for the engagement everyone. Two quick thoughts I have, please refrain from commenting `+1`, we have thumbs up which serve the same purpose! Second, I'm thinking there are two ways of doing this:

1. Have some config that skips certain subpaths (probably the most intuitive)
2. Have a magic hidden file, like `.mockery_skip` that if the walker finds, will skip the entire subdirectory.

I think 2 is a bit more portable and easier to code for. Which of these would people prefer? Or maybe there is a secret third option I'm missing.

alec-w:
I'd vote for option 2, I hadn't even thought of that alternative!

Imagining code to support this would be simpler which is a big plus. Although config might be more intuitive, the exact way in which the subpaths have to be specified would require reading docs more carefully, whereas the `.mockery_skip` usage should be obvious to anyone familiar with things like `.gitignore`.

alec-w:
PR opened for implementing `.mockery_skip`. First time contributing something more than a trivial doc change to OS so any guidance on improving the PR description / approach very welcome 🙂

subtle-byte:
In my mocks generator it is possible to exclude specific subfolder) https://github.com/subtle-byte/mockigo

hedhyw:
**Option 1:**
There are too many configs these days, but this option is more extensible 👍.


**Option 2:**
1. I am not sure about the `.mockery_skip` method, because if we place it in the `/vendor` directory, then it will be deleted after each `go mod vendor` call.
[...]

PR #458, opened 2022-04-26
Support directory exclusion when generating mocks

alec-w:
Description
-------------

Sometimes mocks are wanted for the majority of a project, but specific directories are to be excluded (e.g. `internal`).

This adds support for excluding a directory and its sub-directories from mock generation by placing a file named `.mockery_skip` inside it.

- Fixes #[395](https://github.com/vektra/mockery/issues/395)

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [X] 1.16
- [ ] 1.17
- [ ] 1.18

How Has This Been Tested?
---------------------------

New test added that adds a `.mockery_skip` file into the the fixtures directory and then runs the same walker test as `TestWalkerHere` but asserting that the generated mocks were the reduced set.

Checklist
-----------

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Table of Contents
 - [Expecter Interfaces](#expecter-interfaces)
 - [Mock constructors](#mock-constructors)
 - [Extended Flag Descriptions](#extended-flag-descriptions)
+- [Excluding Directories](#excluding-directories)
 - [Mocking interfaces in `main`](#mocking-interfaces-in-main)
 - [Configuration](#configuration)
   * [Example](#example)
@@ -366,6 +367,11 @@ The following descriptions provide additional elaboration on a few common parame
 | `--print` | Use `mockery --print` to have the resulting code printed out instead of written to disk. |
 | `--exported` | Use `mockery --exported` to generate public mocks for private interfaces. |
 | `--with-expecter` | Use `mockery --with-expecter` to generate `EXPECT()` methods for your mocks. This is the preferred way to setup your mocks. |
+
+Excluding Directories
+---------------------
+
+Directories can be excluded from mock generation by placing a file `.mockery_skip` inside them. Mocks will not be generated for any types present in a directory containing such a file nor for any in its sub-directories.
 
 Mocking interfaces in `main`
 ----------------------------
--- a/pkg/walker.go
+++ b/pkg/walker.go
@@ -68,9 +68,11 @@ func (w *Walker) doWalk(ctx context.Context, p *Parser, dir string, visitor Walk
 
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
+		log.Err(err).Msgf("Error reading directory")
 		return
 	}
 
+OUTER:
 	for _, file := range files {
 		if strings.HasPrefix(file.Name(), ".") || strings.HasPrefix(file.Name(), "_") {
 			continue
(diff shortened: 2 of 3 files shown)


pkg/walker.go
```diff
@@ -80,6 +82,17 @@ func (w *Walker) doWalk(ctx context.Context, p *Parser, dir string, visitor Walk
 
 		if file.IsDir() {
 			if w.Recursive {
+				nextFiles, err := ioutil.ReadDir(path)
+				if err != nil {
+					log.Err(err).Msgf("Error reading directory")
+					return
```
LandonTClipp, line 88: Because the doWalk function doesn't provide a way to bubble the error up, I feel like this should be a fatal error. But, we really should fix `doWalk` not returning errors.

pkg/walker.go
```diff
@@ -80,6 +82,17 @@ func (w *Walker) doWalk(ctx context.Context, p *Parser, dir string, visitor Walk
...
+				}
+				for _, nextFile := range nextFiles {
+					if nextFile.Name() == ".mockery_skip" {
+						continue OUTER
```
LandonTClipp, line 92: I tend to really avoid labels if at all possible. Actually, I think we can `stat` for `.mockery_skip` at the top of `doWalk` and simply return if it exists.

Shameless plug but you might want to look into https://pkg.go.dev/github.com/chigopher/pathlib and then you can just do `pathlib.Path(dir).Joinpath(".mockery_skip").Exists()` to check.

README.md
```diff
@@ -367,6 +368,11 @@ The following descriptions provide additional elaboration on a few common parame
 | `--exported` | Use `mockery --exported` to generate public mocks for private interfaces. |
 | `--with-expecter` | Use `mockery --with-expecter` to generate `EXPECT()` methods for your mocks. This is the preferred way to setup your mocks. |
 
+Excluding Directories
+---------------------
+
+Directories can be excluded from mock generation by placing a file `.mockery_skip` inside them. Mocks will not be generated for any types present in a directory containing such a file nor for any in its sub-directories.
```
LandonTClipp, line 374: Actually, let's call this `.mockery_skip_all`. I envision we might want a `.mockery_skip` that skips _just_ a single directory, whereas the `_all` can skip the entire tree.

alec-w: By "skip the entire tree" vs "just a single directory" is the intended meaning that
- "skip the entire tree" is recursively down the tree from this point
- "just a single directory" is don't generate mocks for files in this directory but do for those in directories beneath this one
? (checking I've got that right in my head)

LandonTClipp: Yep that sounds right, I feel like we might want both options at some point but we can just focus on "skip the entire tree" as you were originally doing. Thanks!

cplane (approved):

LandonTClipp:
@alec-w checking back up on this, I think a number of people would love this feature, do you think you have some time to address the comments? Everyone's busy these days so no worries if not!

LandonTClipp:
For the folks that still want this feature, we will kind of get this "for free" in the feature implemented here: https://github.com/vektra/mockery/pull/548

So I will close this out in favor of that pull request.

PR #466, opened 2022-05-11
Reduce size of interface passed to mock constructor

### LINKED ISSUE #461: Generated Mock Constructors can't be used with Ginkgo
Description
-------------

Generated Mock constructors incompatible with Ginkgo

Version
--------

v2.12.1

Steps to Reproduce
------------------

1. Generate any Mock.
2. Try to instantiate it with:
```
import . "github.com/onsi/ginkgo/v2"

var _ = Describe("fails", func(){
  _ = NewMock(GinkgoT())
})
```

### Expected Behavior

Should build.

### Actual Behavior

Does not build, because `testing.TB` can't be implemented from other packages.

LandonTClipp:
Description
-------------

This PR adds a custom interface that will be used in mock constructors. Previously we had something like this:

```go
// NewExample creates a new instance of Example. It also registers the testing.TB interface on the mock and a cleanup function to assert the mocks expectations.
func NewExample(t testing.TB) *Example {
	mock := &Example{}
	mock.Mock.Test(t)

	return mock
}
```

Now this will be generated:

```go
type NewExampleT interface {
	mock.TestingT
	Cleanup(func())
}

// NewExample creates a new instance of Example. It also registers the testing.TB interface on the mock and a cleanup function to assert the mocks expectations.
func NewExample(t NewExampleT) *Example {
	mock := &Example{}
	mock.Mock.Test(t)

	return mock
}
```

Using `testing.TB` was too broad as many of the methods in the interface were not actually used by the mocks, so we reduce the interface to only the methods actually in use.

- Fixes #461 

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [x] 1.18

How Has This Been Tested?
---------------------------

`go test ./...`

Checklist
-----------


LandonTClipp:
Actually I'm going to modify the comment on the mock constructor as it's wrong (we're not using `testing.TB` anymore). A task for tomorrow.

kgadams:
I was going to mention the comment on the constructor, but you already know about that. Thanks a lot!

cfiderer:
Looks great!

cfiderer:
Thanks a lot for the fast fix!

kgadams:
Thanks a lot from me as well.

gregbarasch:
Just curious -- how is this not a breaking change?

cfiderer:
> Just curious -- how is this not a breaking change?

`testing.TB` is a structure that cannot be implemented outside of the `testing` package.
The interface that is generated now is a subset of the methods from `testing`, it requires only the methods that are actually used (here and by `mock.TestingT`).

So - the requirement to the argument is less strict now.

gregbarasch:
> 

Got it! Thank you for the explanation

gregbarasch:
> > Just curious -- how is this not a breaking change?
> 
> `testing.TB` is a structure that cannot be implemented outside of the `testing` package. The interface that is generated now is a subset of the methods from `testing`, it requires only the methods that are actually used (here and by `mock.TestingT`).
> 
> So - the requirement to the argument is less strict now.

Got it! Thank you for the explanation

PR #498, opened 2022-08-17
Remove named interface

grongor:
Description
-------------

Can't believe I didn't realize we can do this before :) We don't need to create the constructor testing interface for each mock; we can use an anonymous type.

- [x] 1.19

How Has This Been Tested?
---------------------------

Yes :)

Checklist
-----------


LandonTClipp:
D'oh! I don't know if anyone relies on this at this point?

LandonTClipp:
D'oh! I think this should be fine to remove for people who send mocks to the `mocks` directory, but what about people who use `--inpkg`? They might rely on this now.

grongor:
I honestly don't think that they would. I don't see a reason. But yeah, "officially" this is a BC-break. If you wanna play it safe, just merge it into v3 branch :)

PR #509, opened 2022-09-29
Add replace for internal path

### LINKED ISSUE #508: Mocks are not work if generated for `internal`
Description
-------------

For project, where mocks are generated recursively with keeptree flag mocks are not working for `internal`
```
mockery -r --keeptree --all --with-expecter --packageprefix mock
```

Mockery Version
--------

v2.14

Golang Version
--------------

1.19

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [x] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. Create package within internal
2. Generate mocks on top level recursively with `--keeptree`
3. Using generated mocks are not possible

### Expected Behavior

I think when `--keeptree` is used - it shouldn't create internal.

bdandy:
Description
-------------

Replace `internal` path with prefixed `_` for `--keeptree`

- Fixes #508 

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [x] 1.18

How Has This Been Tested?
---------------------------

Checklist
-----------

### DIFF
--- a/pkg/outputter.go
+++ b/pkg/outputter.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/rs/zerolog"
+
 	"github.com/vektra/mockery/v2/pkg/config"
 	"github.com/vektra/mockery/v2/pkg/logging"
 )
@@ -56,6 +57,9 @@ func (p *FileOutputStreamProvider) GetWriter(ctx context.Context, iface *Interfa
 		relativePath := strings.TrimPrefix(
 			filepath.Join(filepath.Dir(iface.FileName), p.filename(caseName)),
 			absOriginalDir)
+
+		relativePath = strings.Replace(relativePath, "/internal/", "/_internal/", -1)
+
 		path = filepath.Join(p.BaseDir, relativePath)
 		if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
 			return nil, err, func() error { return nil }
--- a/pkg/outputter.go
+++ b/pkg/outputter.go
@@ -58,7 +58,8 @@ func (p *FileOutputStreamProvider) GetWriter(ctx context.Context, iface *Interfa
 			filepath.Join(filepath.Dir(iface.FileName), p.filename(caseName)),
 			absOriginalDir)
 
-		relativePath = strings.Replace(relativePath, "/internal/", "/_internal/", -1)
+		// as it's not possible to import from internal path, we have to replace it in mocks when KepTree is used
+		relativePath = strings.Replace(relativePath, "/internal/", "/internal_/", -1)
 
 		path = filepath.Join(p.BaseDir, relativePath)
 		if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {


jippi:
I got hit by this today as well :) 

Would love to see this merged in!

jippi:
@LandonTClipp does the approach suggested here work for adoption into Mockery? if not, I would love to contribute a fix that y'all custodians for Mockery would prefer :)

LandonTClipp:
Hey folks, thanks for the PR. This seems kind of dirty to me, but I do understand the intention and the proposed solution. Could you add a comment to the line describing why this is necessary? Just for future folks who squint at this wondering why we're doing such an odd thing. 

I can't immediately think of a better approach to this at the moment, however.

pkg/outputter.go
```diff
@@ -56,6 +57,9 @@ func (p *FileOutputStreamProvider) GetWriter(ctx context.Context, iface *Interfa
 		relativePath := strings.TrimPrefix(
 			filepath.Join(filepath.Dir(iface.FileName), p.filename(caseName)),
 			absOriginalDir)
+
+		relativePath = strings.Replace(relativePath, "/internal/", "/_internal/", -1)
```
jippi, line 61: Would it work with `/internal_/` too? If yes, I would prefer to keep the lexical sorting in my file explorer intact as much as possible by having the `_` at the end

LandonTClipp:
@bdandy I like @jippi 's suggestion of appending underscore instead of prepending to keep lexical sorting. Do you think you could make that change?

pkg/outputter.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/rs/zerolog"
```
fadhilijuma, line 11: Is there a reason why we use different loggers in this repo? My little understanding tells me that we use Logrus. Why Zerolog again? I propose consistency.

bdandy:
Sounds great, let me add the change

jippi:
Amazing, thank you! :)

jippi:
Nice, thank you for landing this! :) Will make my project layout so much nicer going forward 🎉

jippi:
@LandonTClipp I'm not sure if this is a bug in the new `packages` config or not, but `internal` is not being replaced with `internal_`  when using a config like this

```yaml
packages:
  ./internal:
    config:
      all: true
      recursive: true
      #inpackage: true
      disable-version-string: true
      dir: "mocks/{{ .InterfaceDirRelative }}_mocks"
      filename: "{{ .InterfaceName }}.go"
      mockname: "{{ .Mock }}{{ .InterfaceName }}"
      outpkg: "{{ .PackageName }}_mocks"
```

jippi:
@bdandy @LandonTClipp  I've spun up #639 which I think is a pretty nice (and flexible!) solution to this problem :)

LandonTClipp:
I don't think we could revert this as people might be relying on this. The changes in this PR were implemented for the legacy configuration scheme, which I intend to delete anyway in v3. We can't backport your functionality because it requires using the templated variables, which are not supported in the legacy scheme (it's two entirely different code paths and I don't want to bother backporting it into code that will die anyway).

PR #538, opened 2023-02-04
Implement proposed changes to generator

### LINKED ISSUE #530: mockery should allow returning values based on some custom logic
Description
-------------
In the expect mode, in addition to the Run method there should be a RunAndReturn method which allows users to execute custom logic and then return data based on the logic

Mockery Version
--------

version=v2.16.0

Golang Version
--------------

go1.19.4

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [ X ] go install
- [ ] Other: [specify]

### Expected Behavior
I have an interface 

```go
type Executor interface {
	Execute(name string) error
}
```

```go
func Test_Executor(t *testing.T) {
	executor := mocks.NewExecutor(t)
	executor.EXPECT().Execute(mock.Anything).RunAndReturn(func(name string) error {
		if name == "foo" {
			return nil
		}
		return errors.New("invalid data")
	})
	
	executor.Execute("foo") // should return nil
	executor.Execute("bar") // should return errors.New("invalid data")
	
}
```
Thanks....

### LINKED ISSUE #537: Suggested improvement to "Return Value Provider Functions" feature
Mockery's generated code currently allows you to pass in functions as the arguments to Return(), where those functions accept the same parameters as the function being called, and the return value is _one_ of the return parameters of the function.  This can get a little clunky when mocking functions with multiple outputs, as you need to define one function for each output.

If the generated code were tweaked slightly, it can support passing in a single function to Return() (regardless of the number of return parameters), where that function would be one that matches the exact signature of the mock being called.

```go
// given this input:

//go:generate mockery --name MyInterface
type MyInterface interface {
	TestMethod(a, b int) (int, int)
}

// suggestion: generate this code:

// TestMethod provides a mock function with given fields: a, b
func (_m *MyInterface) TestMethod(a int, b int) (int, int) {
	ret := _m.Called(a, b)

	var r0 int
	var r1 int

	if rf, ok := ret.Get(0).(func(int, int) (int, int)); ok {
		r0, r1 = rf(a, b)
	} else {
		if rf, ok := ret.Get(0).(func(int, int) int); ok {
			r0 = rf(a, b)
		} else {
			r0 = ret.Get(0).(int)
		}

		if rf, ok := ret.Get(1).(func(int, int) int); ok {
			r1 = rf(a, b)
		} else {
			r1 = ret.Get(1).(int)
		}
	}

	return r0, r1
}
```

All that was needed was to move the declaration of the return parameters r0 and r1 up to the top of the function, add one extra type conversion just after that, and put everything else that mockery currently does into an `else` block. 

This is backwards-compatible with the existing offering:

image

dlwyatt:
Description
-------------

Allows a single provider function to be used which provides all of the return parameters for a mocked function, instead of needing to define a separate function per output.  The older functionality still works as well.

Also adds a new RunAndReturn function to expecter calls to use this same type of provider, functionally equivalent to `.Call.Return(providerFunc)`

- Fixes #537 
- Fixes #530

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [x] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

Updated generator_test.go and ran `make test`.  Was not able to run `make all` on my system, something about the `run-e2e.sh` script didn't work for me.  (I'm running Ubuntu in WSL on Windows 11, if it matters.)

I also ran the new build against some of my own test code, such as the samples in #537 , to make sure the generated code behaved as expected when `go test` was run.

Checklist
-----------

### DIFF
--- a/mocks/pkg/fixtures/Fooer.go
+++ b/mocks/pkg/fixtures/Fooer.go
@@ -19,7 +19,6 @@ func (_m *Fooer) Baz(path string) func(string) string {
 	ret := _m.Called(path)
 
 	var r0 func(string) string
-
 	if rf, ok := ret.Get(0).(func(string) func(string) string); ok {
 		r0 = rf(path)
 	} else {
@@ -36,7 +35,6 @@ func (_m *Fooer) Foo(f func(string) string) error {
 	ret := _m.Called(f)
 
 	var r0 error
-
 	if rf, ok := ret.Get(0).(func(func(string) string) error); ok {
 		r0 = rf(f)
 	} else {
--- a/mocks/pkg/fixtures/Fooer.go
+++ b/mocks/pkg/fixtures/Fooer.go
@@ -19,6 +19,7 @@ func (_m *Fooer) Baz(path string) func(string) string {
 	ret := _m.Called(path)
 
 	var r0 func(string) string
+
 	if rf, ok := ret.Get(0).(func(string) func(string) string); ok {
 		r0 = rf(path)
 	} else {
@@ -35,6 +36,7 @@ func (_m *Fooer) Foo(f func(string) string) error {
 	ret := _m.Called(f)
 
 	var r0 error
+
 	if rf, ok := ret.Get(0).(func(func(string) string) error); ok {
 		r0 = rf(f)
 	} else {
--- a/pkg/generator.go
+++ b/pkg/generator.go
@@ -816,7 +816,7 @@ func {{ .ConstructorName }}{{ .TypeConstraint }}(t {{ .ConstructorTestingInterfa
 	g.printTemplate(data, constructorTemplate)
 }
 
-// generateCalled returns the Mock.Called invocation string and, if necessary, prints the
+// generateCalled returns the Mock.Called invocation string and, if necessary, a preamble with the
 // steps to prepare its argument list.
 //
 // It is separate from Generate to avoid cyclomatic complexity through early return statements.
--- a/pkg/generator.go
+++ b/pkg/generator.go
@@ -313,16 +313,20 @@ func (g *Generator) printf(s string, vals ...interface{}) {
 var templates = template.New("base template")
 
 func (g *Generator) printTemplate(data interface{}, templateString string) {
-	err := templates.ExecuteTemplate(&g.buf, templateString, data)
+	tmpl := templates.New(templateString).Funcs(
+		template.FuncMap{
+			"join": strings.Join,
+		},
+	)
+
+	tmpl, err := tmpl.Parse(templateString)
 	if err != nil {
-		tmpl, err := templates.New(templateString).Parse(templateString)
-		if err != nil {
-			// couldn't compile template
-			panic(err)
-		}
-		if err := tmpl.Execute(&g.buf, data); err != nil {
-			panic(err)
-		}
+		// couldn't compile template
+		panic(err)
+	}
+
+	if err := tmpl.Execute(&g.buf, data); err != nil {
+		panic(err)
 	}
 }
 
@@ -471,6 +475,32 @@ type paramList struct {
 	Variadic   bool
 }
 
+func (p *paramList) FormattedParamNames() string {
+	formattedParamNames := ""
+	for i, name := range p.Names {
+		if i > 0 {
+			formattedParamNames += ", "
+		}
+
+		paramType := p.Types[i]
+		// for variable args, move the ... to the end.
+		if strings.Index(paramType, "...") == 0 {
+			name += "..."
+		}
+		formattedParamNames += name
+	}
+
+	return formattedParamNames
+}
+
+func (p *paramList) ReturnNames() []string {
+	var names = make([]string, 0, len(p.Names))
+	for i := 0; i < len(p.Names); i++ {
+		names = append(names, fmt.Sprintf("r%d", i))
+	}
+	return names
+}
+
 func (g *Generator) genList(ctx context.Context, list *types.Tuple, variadic bool) *paramList {
 	var params paramList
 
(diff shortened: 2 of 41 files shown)


dlwyatt:
Note on the checklist; the first item says "My code follows the style guidelines of this project", but I haven't actually found any style guidelines in the repo.  I tried to match the style of the code around where I made modifications.

I also didn't add any comments, as the sections of code that I was modifying already didn't contain any.  I can add some comments around the new bits if you prefer.

denizgursoy:
@dlwyatt Do you know whether it is going to be merged or know someone who we can ask if they are going to merge? Should we ask it to @LandonTClipp?

dlwyatt:
No idea. 🙂 This is my first interaction on this project.

LandonTClipp:
Hi folks, thanks for the contribution. I will have to review the proposed change over the coming days. I work a full-time job so I can't commit to a specific period but I do try my best so I hope to have at least a first pass review by the end of this week :)

mocks/pkg/fixtures/Fooer.go
```diff
@@ -19,6 +19,7 @@ func (_m *Fooer) Baz(path string) func(string) string {
 	ret := _m.Called(path)
 
 	var r0 func(string) string
+
```
LandonTClipp, line 22: Could we get rid of whatever change causes this whitespace?

dlwyatt: Sure, no problem

pkg/generator.go
```diff
@@ -611,26 +611,45 @@ func (g *Generator) Generate(ctx context.Context) error {
...
+
+			for idx, typ := range returns.Types {
+				if idx > 0 {
+					g.printf("\n")
+				}
+
+				g.printf("%s\tif rf, ok := %s.Get(%d).(func(%s) %s); ok {\n",
+					tab, retVariable, idx, strings.Join(params.Types, ", "), typ)
+				g.printf("%s\t\tr%d = rf(%s)\n", tab, idx, formattedParamNames)
+				g.printf("%s\t} else {\n", tab)
```
LandonTClipp, line 635: You know a part of me wants to stop using individual printf statements because it's difficult for me to understand what's going on. You have the right approach of trying to follow what the code historically did, but I think it'd be better for readability to use the `printTemplate` method instead. What do you think? Is that easily done or is printf a better fit here?

dlwyatt: I don't find either of them all that easy to follow, tbh.  It's easier to look at the expected outputs in generator_test.go, and trust the tests to make sure all the templates and printf's are actually producing that.

dlwyatt: Comparing before and after on one of the test cases, this PR changes this:

```go
// Put provides a mock function with given fields: path
func (_m *RequesterReturnElided) Put(path string) (int, error) {
	ret := _m.Called(path)

	var r0 int
	if rf, ok := ret.Get(0).(func(string) int); ok {
		r0 = rf(path)
	} else {
		r0 = ret.Get(0).(int)
	}

	var r1 error
	if rf, ok := ret.Get(1).(func(string) error); ok {
		r1 = rf(path)
	} else {
		r1 = ret.Error(1)
	}

	return r0, r1
}
```

into this:

```go
// Put provides a mock function with given fields: path
func (_m *RequesterReturnElided) Put(path string) (int, error) {
	ret := _m.Called(path)

	var r0 int
	var r1 error
	if rf, ok := ret.Get(0).(func(string) (int, error)); ok {
		r0, r1 = rf(path)
	} else {
		if rf, ok := ret.Get(0).(func(string) int); ok {
			r0 = rf(path)
		} else {
			r0 = ret.Get(0).(int)
		}

		if rf, ok := ret.Get(1).(func(string) error); ok {
			r1 = rf(path)
		} else {
			r1 = ret.Error(1)
		}
	}

	return r0, r1
}
```

And adds this to the expecter calls:

```go
func (_c *RequesterReturnElided_Put_Call) RunAndReturn(run func(string) (int, error)) *RequesterReturnElided_Put_Call {
	_c.Call.Return(run)
	return _c
}
```

dlwyatt: I can try, but that's a heck of a lot more work to rewrite more of the code than what I've modified so far, and getting templates to produce exactly the same output might be pretty annoying.

dlwyatt: This was a _giant_ pain in the butt, and I did have to tweak the new output slightly, but I managed to get a template to generate the same code that the printf stuff was doing.  (Technically `generateCalled` is sort of cheating and just using `fmt.Sprintf` instead of `g.printf` in a few places, but it'll do for now.)

The thing that I changed in the expected outputs is to do this:

```go
	if rf, ok := ret.Get(0).(func(string, uint16) ([]byte, *test.Err)); ok {
		return rf(_a0, _a1)
	}
	if rf, ok := ret.Get(0).(func(string, uint16) []byte); ok {
		r0 = rf(_a0, _a1)
	} else {
		if ret.Get(0) != nil {
			r0 = ret.Get(0).([]byte)
		}
	}

	if rf, ok := ret.Get(1).(func(string, uint16) *test.Err); ok {
		r1 = rf(_a0, _a1)
	} else {
		if ret.Get(1) != nil {
			r1 = ret.Get(1).(*test.Err)
		}
	}

	return r0, r1
```

Note the early return if the new type of provider function is found.  This way there doesn't need to be an `else` block, and the indentation of the rest of the function is much easier to deal with.

dlwyatt: (Don't even ask me what's going on with all the `{{- }}` and `{{ -}}` bits in the new template parts.  I was just poking at it until all the whitespace exactly matched what was already in the test expectations.

LandonTClipp: This is great! Sorry, if I knew how much of a pain it would be maybe I could have let it slide but I do really enjoy the templates a lot more. At least to me, it's easier to reason about.

I'll ack this, lmk if there are any finishing touches you want to make before I merge.

LandonTClipp (approved):

pkg/generator.go
dlwyatt: I need to walk away for a bit and look at it with fresh eyes later, maybe in the morning.  I'll let you know if I spot anything that could use work then.

LandonTClipp: Sounds good, thanks for the stellar work. It's really appreciated.

dlwyatt: I think this is good to go.  There was one function comment that was sort of inaccurate after I changed things, but I think the code itself is okay.

LandonTClipp: Great work, I’ll take one last look tonight and merge if everything is good. Thanks Dave.

LandonTClipp (approved):

PR #550, opened 2023-02-12
Variadic Method Improvements (Work In Progress - Do Not Merge)

### LINKED ISSUE #541: Suggested improvement around variadic functions
I've been looking over mockery's history behind this topic.  Back in 2017, there was #123 , which introduced behavior to "unroll" variadic arguments when calls were made to the mock.  You can read the PR for more details, but the gist of it was, people wanted to be able to write `mock.On("MethodName", 1,2,3,4,5)` instead of `mock.On("MethodName", []interface{}{1,2,3,4,5})`.  The change in that PR did allow for that syntax to work, but the problem was that it broke the ability to pass a single value of `mock.Anything` to cover any number of arguments to the variadic parameter.

In #359, the `--unroll-variadic=false` flag was added to restore the original behavior, for people who really wanted that single `mock.Anything` to work, and didn't care about losing the other syntax.

What I'm proposing here may be a "best of both worlds" approach.  The tests I'll list here are all working, and if anyone can think of other use cases that I haven't thrown at it yet, I'll include those as well.  Essentially it inverts the approach used in #123 ; instead of "unrolling" the variadic arguments when the mocked function is called, instead "roll them up" in new wrappers around the `On`, `AssertCalled` and `AssertNotCalled` methods of a testify Mock.  This means that you still get the nice syntax sugar of `mock.On("Method", 1,2,3,4,5)` without having to put things into slices yourself, and a single `mock.Anything` parameter will still match any number (zero or more) arguments passed to the variadic parameter.  Here's the interface I worked against:

```go
type SampleInterface interface {
	Sprintf(msg string, args ...interface{}) string
	Test(msg string) string
}
```

I deliberately included one non-variadic method signature in there to make sure those still work properly.  My test code looks like this:

```go
func TestVariadicWrappers(t *testing.T) {
	m := new(mocks.SampleInterface)
	m.On("Sprintf", "specific", "arguments").Return("Boo!")
	m.On("Sprintf", mock.Anything, mock.Anything).Return("Yay!")

	// deliberately passing in more than one arguments to the variadic parameter
	// this is what fails today, even though we mocked with mock.Anything
	s := m.Sprintf("I am a %s %s", "test", "two")
	assert.Equal(t, "Yay!", s)

	// Making sure things still work correctly with a non-variadic signature
	m.On("Test", "something").Return("Yay!")
	s = m.Test("something")
	assert.Equal(t, "Yay!", s)

	m.AssertCalled(t, "Test", "something")
	m.AssertCalled(t, "Sprintf", "I am a %s %s", "test", "two")

	// Also making sure calling with ZERO arguments to the variadic parameter works
	m.Sprintf("bagel")
	m.AssertCalled(t, "Sprintf", "bagel")

	m.AssertNotCalled(t, "Sprintf", "bagel", "with", "cream", "cheese")
	m.AssertNotCalled(t, "Sprintf", "donut")

	// Making sure the matchers are still working when we're not using mock.Anything
	s = m.Sprintf("specific", "arguments")
	assert.Equal(t, "Boo!", s)
}
```

You can see how I'm running through the paces of both variadic and non-variadic method signatures, and mocks that use mock.Anything or pass other values in.  Mocks generated with today's code (whether `--unroll-variadic` was set to true or false) will all fail.  But with this code, the tests all pass:

```go
package mocks

import (
	"reflect"

)

// SampleInterface is an autogenerated mock type for the SampleInterface type
type SampleInterface struct {
	mock.Mock
}

// Sprintf provides a mock function with given fields: msg, args
func (_m *SampleInterface) Sprintf(msg string, args ...interface{}) string {
	ret := _m.Called(msg, args)

	var r0 string

	if rf, ok := ret.Get(0).(func(string, ...interface{}) string); ok {
		r0 = rf(msg, args...)
	} else {
		if rf, ok := ret.Get(0).(func(string, ...interface{}) string); ok {
			r0 = rf(msg, args...)
		} else {
			r0 = ret.Get(0).(string)
		}
	}

	return r0
}

// Test provides a mock function with given fields: msg
func (_m *SampleInterface) Test(msg string) string {
	ret := _m.Called(msg)

	var r0 string

	if rf, ok := ret.Get(0).(func(string) string); ok {
		r0 = rf(msg)
	} else {
		if rf, ok := ret.Get(0).(func(string) string); ok {
			r0 = rf(msg)
		} else {
			r0 = ret.Get(0).(string)
		}
	}

	return r0
}

func (_m *SampleInterface) On(methodName string, arguments ...interface{}) *mock.Call {
	arguments = _m.rollVariadic(methodName, arguments...)
	return _m.Mock.On(methodName, arguments...)
}

func (_m *SampleInterface) AssertCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
	arguments = _m.rollVariadic(methodName, arguments...)
	return _m.Mock.AssertCalled(t, methodName, arguments...)
}

func (_m *SampleInterface) AssertNotCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool {
	arguments = _m.rollVariadic(methodName, arguments...)
	return _m.Mock.AssertNotCalled(t, methodName, arguments...)
}

func (_m *SampleInterface) rollVariadic(methodName string, arguments ...interface{}) []interface{} {
	sig := _m.getMethodSignature(methodName)

	if !sig.IsVariadic() {
		return arguments
	}

	variadicIndex := sig.NumIn() - 1
	if len(arguments) == sig.NumIn() && arguments[variadicIndex] == mock.Anything {
		return arguments
	}

	newArgs := make([]interface{}, sig.NumIn())

	copy(newArgs, arguments[0:variadicIndex])

	if len(arguments) >= sig.NumIn() {
		newArgs[variadicIndex] = arguments[variadicIndex:]
	} else {
		newArgs[variadicIndex] = []interface{}(nil)
	}

	return newArgs
}

func (_m *SampleInterface) getMethodSignature(methodName string) reflect.Type {
	switch methodName {
	case "Sprintf":
		return reflect.TypeOf(_m.Sprintf)
	case "Test":
		return reflect.TypeOf(_m.Test)
	default:
		panic("Invalid method name")
	}
}

type mockConstructorTestingTNewSampleInterface interface {
	mock.TestingT
	Cleanup(func())
}

// NewSampleInterface creates a new instance of SampleInterface. It also registers a testing interface on the mock and a cleanup function to assert the mocks expectations.
func NewSampleInterface(t mockConstructorTestingTNewSampleInterface) *SampleInterface {
	mock := &SampleInterface{}
	mock.Mock.Test(t)

	return mock
}
```

The significant changes from what mockery generates today are:
- The actual calls to the mocked functions behave as though `--unroll-variadic` were set to false; the arguments are passed straight on to `mock.Called`
- New wrappers are added to shadow calls to `On`, `AssertCalled`, and `AssertNotCalled`.  These all call the underlying testify methods after first calling a new `rollVariadic` helper function.
- The `rollVariadic` function itself is static and would be the same in every generated mock file.
- A new getMethodSignature helper function is there to help `rollVariadic` know about the expected parameters associated with each method name.  This function is dynamic, but easily generated based only on a list of method names in the mock.

If you like this approach, I'm happy to create a PR that will update the generator to actually produce this new code.  The `--unroll-variadic` flag would be deprecated and have no further effect on the output, at that point.

dlwyatt:
Description
-------------

Proposed improvements for mocking variadic methods.  Moving the dialogue from #541 over here so further discussion can be made alongside visible code diffs.

Note - In order to cut down on or eliminate the number of breaking changes this code would introduce, I have a PR open in testify here:  https://github.com/stretchr/testify/pull/1348 .  I don't know if they will take that PR yet.  For now, this PR points at my fork of testify to demonstrate how most of the old tests are working with the new generated code.

Lots more details in the linked issue, but the short version here is that this PR aims to get rid of the --unroll-variadic flag, and have mockery give us the best of both worlds from how it works today when that flag is set to either true or false.  Namely, that you can use mock's `On`, `AssertCalled`, and `AssertNotCalled` functions without manually building slice values for the variadic parameters, and you can still use a single `mock.Anything` value against variadic parameters to match any number of arguments.

Assuming testify takes the PR, there is still one breaking change remaining in the existing code; you _can't_ manually build up a slice value for the variadic parameter anymore, which was a syntax still being tested by certain parts of generator_test.go (around line 1500-1600).  It may be possible to still allow that syntax in most cases by having the generated mocks look for situations where a single expected argument was passed for a variadic parameter, and that argument is itself a slice.  In that case, pass the slice on unmodified to testify.  (This might still cause the odd corner case to fail, when someone really does want a slice containing exactly one other slice for some reason.  Hopefully there's not too much Go code in the world that looks that way.  😄 )

- Fixes #541 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [x] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

Run `make test`

Checklist
-----------


go.mod
```diff
@@ -40,3 +40,5 @@ require (
 	gopkg.in/ini.v1 v1.66.6 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/stretchr/testify => github.com/dlwyatt/testify v1.9.9
```
dlwyatt, line 44: temporary until we see if testify takes the linked PR.

LandonTClipp:
You already touch on this but I want to get some clarity on this specific question. Take this production code:

```go
func Foo(s any...) {}

// case 1
Foo("foo", "bar")
// case 2
Foo("foo")
```

In the tests I may have this:

```go
mockIntf.On("Foo", mock.Anything)
```

I may either:
1. Want to assert _any number_ of variadic args of any value, or
2. Want to assert a single argument in the variadic parameter of any value

It's not clear to me in the proposal how one would differentiate between those two cases.

In my opinion, it seemed like it was a mistake allowing the syntactic sugar because now these cases I mentioned are ambiguous. With `unroll-variadic: False` and your PR to testify merged, this becomes simple:

```
// Asserts any number of arguments of any value
mockIntf.On("Foo", mock.Anything)
```

```
// Asserts a single argument of any value
mockIntf.On("Foo", []any{mock.Anything})
```

dlwyatt:
"now these cases I mentioned are ambiguous"

That's an interesting point. I can see a few possibilities here for matching exactly one argument, assuming the testify PR gets merged:

- To match exactly one argument, use AnythingOfType
- Make the update I mentioned earlier where you can still pass in a slice of expected values, and have that slice contain a single element of mock.Anything
- Add a new placeholder that is specific to Mockery (VariadicAnything or something like that), which conveys the "match any number of arguments" meaning, while mock.Anything remains used for matching a single argument.

LandonTClipp:
> To match exactly one argument, use AnythingOfType

This feels inconsistent and will be a weird edge case.

> Make the update I mentioned earlier where you can still pass in a slice of expected values, and have that slice contain a single element of mock.Anything

This is the best option IMO. But then like you mentioned before, you get into a situation where someone might be passing in a slice to the mock implementation as the first argument to the variadic. Kind of weird but... it could happen!

> Add a new placeholder that is specific to Mockery (VariadicAnything or something like that), which conveys the "match any number of arguments" meaning, while mock.Anything remains used for matching a single argument.

It's an interesting thought. I would rather not start maintaining our own sentinels as it's this weird edge case wart. In my opinion it really seems like the way that solves all of these problems is getting your testify PR merged and making `unroll-variadic: False`. No ambiguity at all. Some people may be unhappy with the lack of syntactic sugar but as long as we clearly identify this case in the docs, I don't think it would be an issue. Other solutions to me feel like bandaids and open up lots of edge cases we couldn't account for.

dlwyatt:
Oh, one other option. Make a custom type alias for []any for when you want to be explicit about the slice. If you pass in a single arg of that type, mockery knows to pass it unmodified to testify. Otherwise, roll it up. That covers all the bases (for these examples I'm using VarSlice as the made up type name):

- mock any number of arguments, pass mock.Anything
- mock a single argument only, but don't care about its value, pass in VarSlice{mock.Anything}
- Want to pass in a single argument that happens to be another slice, just pass in your []whatever value. It's not a VarSlice so it'll get wrapped up in a single element []any for testify.

This does still introduce some new usage and could break existing tests, but handles it all in non ambiguous ways without losing the variadic style syntax for most use cases.

dlwyatt:
Come to think of it, this could be done on a different way that doesn't require a change to testify. Leave the old unroll behavior mostly as is, and just add a new placeholder to mockery for the "match any number of arguments" case. Old tests continue to work fine. Benefit of --unroll-variadic=false is available with the nicer syntax.

LandonTClipp:
I am inclined to not fix this and make `unroll-variadic: True` deprecated. I think your ideas of having mockery-specific sentinels would definitely _work_ but from a stylistic point of view, I want to keep our call signatures as close to testify as possible in this situation. I think the cost of having less-nice syntax is marginal when compared to the maintenance burden of having more sentinels and additional logic to handle them.

I think there are merits to the argument either way but my personal preference is to not "fix" this in mockery, and just deprecate `unroll-variadic: True`.

LandonTClipp:
@dlwyatt I'm reaching out to the testify maintainers directly to see if we can get your PR merged.

PR #596, opened 2023-04-09
Exclude subdirectories and files with --exclude

### LINKED ISSUE #395: Exclude specific subfolder from --all

amirbenun:
Description
-------------

Add `--exclude` CLI parameter to exclude subdirectories when using `--all`.
The flag `--exclude` can be used more than once on the same call, for example:
```
mockery --dir . --inpackage --all --with-expecter --case underscore --recursive --exclude vendor --exclude resources/providers/
```

- Fixes #395 

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [x] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

I built the binary, copied it to another project, and generated mocks using the command:
```
./mockery --dir . --inpackage --all --with-expecter --case underscore --recursive
```
I tried to exclude different paths in the project and used the new flag multiple times to make sure all paths were excluded.

Checklist
-----------

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -556,6 +556,135 @@ func TestConfig_GetPackages(t *testing.T) {
 	}
 }
 
+func TestConfig_ShouldGenerateInterface(t *testing.T) {
+	tests := []struct {
+		name    string
+		c       *Config
+		want    bool
+		wantErr bool
+	}{
+		{
+			name: "no packages return error",
+			c: &Config{
+				Packages: map[string]interface{}{},
+			},
+			want:    false,
+			wantErr: true,
+		},
+		{
+			name: "should generate all interfaces",
+			c: &Config{
+				Packages: map[string]interface{}{
+					"some_package": map[string]interface{}{},
+				},
+				All: true,
+			},
+			want: true,
+		},
+		{
+			name: "should generate this package",
+			c: &Config{
+				Packages: map[string]interface{}{
+					"some_package": map[string]interface{}{
+						"config": map[string]interface{}{
+							"all": true,
+						},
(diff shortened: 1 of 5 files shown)


LandonTClipp:
Thanks for the contribution. On first glance this is basically exactly how I’d implement it. I’ll give it a more thorough look hopefully later today.

LandonTClipp (approved):
I have zero complaints with this, good job. If you can squash the commits down to one then I'll get it merged and built. Thanks!

pkg/config/config_test.go
```diff
@@ -556,6 +556,135 @@ func TestConfig_GetPackages(t *testing.T) {
 	}
 }
 
+func TestConfig_ShouldGenerateInterface(t *testing.T) {
```
LandonTClipp, line 559: Thanks for adding this test. I had integration tests that were checking this did the right thing but good to have explicit unit test as well 👍🏻

amirbenun:
Thanks @LandonTClipp, squashed it down to a single commit.

pkg/config/config_test.go
carlos2martinize: It gives security if you need to edit guides and options for a library utility

PR #629, opened 2023-05-23
Add support for `unrolled-variadics=false` with expecter structs

dillonstreator:
Description
-------------

- Fixes # (issue)

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [ ] 1.20

How Has This Been Tested?
---------------------------

Checklist
-----------


pkg/generator_test.go
```diff
@@ -559,15 +559,285 @@ func NewExpecter(t mockConstructorTestingTNewExpecter) *Expecter {
...
+func NewExpecter(t mockConstructorTestingTNewExpecter) *Expecter {
+	mock := &Expecter{}
+	mock.Mock.Test(t)
+
+	t.Cleanup(func() { mock.AssertExpectations(t) })
+
+	return mock
+}
+`
+
```
LandonTClipp, line 834: I am trying to move away from the tests asserting the exact string of the generated mock as we currently have to modify like 40 different tests whenever the mock generation code changes. Instead, the better way to do this is add the mock with the specified config into the `.mockery.yaml` file, and run tests off of that mock: https://github.com/vektra/mockery/blob/master/.mockery.yaml#L23

You'd need to add another section under the `configs` of `RequesterVariadic`, something like:
```yaml
- mockname: RequesterVariadicExpecterAndUnrollVariadic
  unroll-variadic: false
  with-expected: true
```

It should be good enough for now to commit the mock and run the test. I'm going to implement a change to CI/CD so that it generates the mocks _before_ running tests (it currently just generates them after).

dillonstreator: Adjusted to follow this testing pattern. I also cleaned up other tests to use the `require` interface and removed the incorrect messages.

pkg/generator_test.go
```diff
@@ -59,7 +59,7 @@ func (s *GeneratorSuite) checkGenerationWithConfig(
 ) *Generator {
 	generator := s.getGeneratorWithConfig(filepath, interfaceName, cfg)
 	err := generator.Generate(s.ctx)
-	s.NoError(err, "The generator ran without errors.")
+	s.Require().NoError(err)
```
LandonTClipp, line 62: Great!

PR #641, opened 2023-06-14
Chore: Update version in the docs.

CoolGoose:
Description
-------------

## Type of change

- [x] This is a documentation update

How Has This Been Tested?
---------------------------

Just does a simple bump in the version from the install.md . I didn't check if there's a way we could maybe pull the version automatically like in the top right side of the website,

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,7 +14,7 @@ Visit the [releases page](https://github.com/vektra/mockery/releases) to downloa
 
 Supported, but not recommended: [see wiki page](https://github.com/vektra/mockery/wiki/Installation-Methods#go-install) and [related discussions](https://github.com/vektra/mockery/pull/456).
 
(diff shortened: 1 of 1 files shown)


LandonTClipp:
When you inspect our mkdocs-material page, the x-path of the git tag element is `/html/body/header/nav/div[3]/a/div[2]/ul/li[1]/text()`

The typescript that manages this is: https://github.com/squidfunk/mkdocs-material/blob/0c45c6fd70a4c17a4647135384ca13e6d4b5efb5/src/assets/javascripts/templates/source/index.tsx#L41

It seems to be using react. So we could probably make another typescript function (or reuse what exists) that returns just the tag name. That would be a good issue for a webdev if anyone wants to give it a shot :)

CoolGoose:
Cool, I'll give it a try on Tuesday evening :)

PR #693, opened 2023-08-16
[Fix] Generate mock build constraints

### LINKED ISSUE #691: tags are not applied to generated mocks
Description
-------------

Tags are not applied to generated mocks.

Mockery Version
--------

```sh
$ mockery --version
13 Aug 23 10:00 CEST INF Starting mockery dry-run=false version=v2.32.4
13 Aug 23 10:00 CEST INF Using config: /projects/poc-mockery/.mockery.yaml dry-run=false version=v2.32.4
v2.32.4
```

Golang Version
--------------

```sh
$ go version
go version go1.21.0 darwin/arm64
```

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [X] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. Create config similar to below:
```yaml
tags: "mock"
packages:
  poc-mockery/pkg/services/oauthstorage:
    interfaces:
      Storage:
```

2. Generate mocks with `mockery` command (no args)

### Expected Behavior

Expected generated file to have `mock`  build tags

### Actual Behavior

No build tags are added. Just the `Code generated....` header

mcdoker18:
Supported only the new format of build tags
https://pkg.go.dev/cmd/go#hdr-Build_constraints

Fixes #691

Description
-------------

- Fixes # (issue)

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

Unit and e2e tests are provided.

Checklist
-----------


mcdoker18:
@LandonTClipp 

There is a little problem with the tag field in config. It is already being used with [parser](https://github.com/vektra/mockery/blob/8abb7022eefaf0d424283d8957117e31e6eaa1cf/pkg/parse.go#L51).  By the way it's not documented behavior. And the worst thing is, the format differs for the parser and file. 
For example:
* parser - `custom1,custom2`
* file - `custom1 && custom2`

LandonTClipp:
I haven't forgotten about this PR, it's on my radar, just been busy with other priorities :)

mocks/github.com/vektra/mockery/v2/pkg/fixtures/buildtag/comment/IfaceWithBuildTagInComment.go
```diff
@@ -0,0 +1,87 @@
+//go:build custom && (!windows || !linux || !darwin || !freebsd)
+
+// Code generated by mockery. DO NOT EDIT.
```
LandonTClipp, line 3: We need to double check if this `DO NOT EDIT` comment being placed below the first line still meets the specification: https://github.com/golang/go/issues/41196

cmd/mockery.go
```diff
@@ -381,6 +381,7 @@ func (r *RootApp) Run() error {
 		InPackage:            r.Config.InPackage,
 		KeepTree:             r.Config.KeepTree,
 		Note:                 r.Config.Note,
+		BuildTags:            r.Config.BuildTags,
```
LandonTClipp, line 384: I have (at least) two questions.

1. So, I'm confused about this line. Is this PR attempting to fix this behavior for both `packages` and non-`packages` configurations? Or are you intending to fix it for one or the other? 

2. My other question, is this okay for us to be passing the build tags into `packages.Load`, and to inject those same build tags as a comment into the mock file? My intuition says yes, but I fear there might be a reason why someone would want different tags for the package parsing, and another for the mock file. But I just can't foresee what would be desirable here.

LandonTClipp:
@mcdoker18 I have a few questions, lmk your thoughts

LandonTClipp:
That would be great if you could push this further! I’ll be sure to get it
merged as soon as my questions are addressed.

Regards,
--
*Landon Clipp*

LandonTClipp:
Hi @wobondar . I'm happy to merge this but I need the discussions to be resolved. I'm also happy to take a different PR as this one seems to be stale. 

I might end up doing this myself at some point as multiple people have requested this.

mocks/github.com/vektra/mockery/v2/pkg/fixtures/buildtag/comment/IfaceWithBuildTagInComment.go
mcdoker18: Thanks for pointing it out, I will fix it.

mcdoker18:
@LandonTClipp I am very sorry. Somehow I skipped notifications from this PR :(

cmd/mockery.go
mcdoker18: 1. What is non-packages configuration? Do you mean default top-level configuration?
2. No, I think that is wrong. Because they have different syntax:
* for generated mocks - https://pkg.go.dev/cmd/go#hdr-Build_constraints
* for `packages.Load` - a comma-separated list

Now `tag` config field is used for loading packages but documentation says for generated mocks.

Maybe we need two different config options for generated mocks and package loading?

LandonTClipp: > What is non-packages configuration? Do you mean default top-level configuration?

I mean usage of the legacy configuration semantics, i.e. things that do not use [this](https://vektra.github.io/mockery/latest/features/#packages-configuration).

> Now tag config field is used for loading packages but documentation says for generated mocks. Maybe we need two different config options for generated mocks and package loading?

Yeah, I am thinking this is what we ultimately want to do. Why don't we go ahead and add two new config options:

```
PackagesLoadTags
MockBuildTags
```

I think that should make it clear enough what is intended?

LandonTClipp: To make this more clear, as I had to remind myself too, the current `build-tags` field gets added to the configuration of the [`pkg.Parser` object](https://github.com/vektra/mockery/blob/v2.39.1/pkg/parse.go#L50-L52). This ultimately makes its way to `packages.Load` [here](https://github.com/vektra/mockery/blob/v2.39.1/pkg/parse.go#L73). So `BuildTags` is specifically meant for _only_ `packages.Load`.

What we want, in addition to what we already have, is `mock-build-tags` which is what will ultimately get added as the `//go:build` directive. So this PR is incorrect in the sense that it's sending the `BuildTags` to the `//go:build` string, when it should only go to the `packages.Load` invocation.

mcdoker18: Sounds good! I intend to add these fields only for the new configuration.

But what is about the current `tags` field? Should I leave it in the source code to be backward compatible and remove from the documentation?

LandonTClipp:
> @LandonTClipp I am very sorry. Somehow I skipped notifications from this PR :(

No worries, thanks for checking back up! If you don't have the time to finish this out that's okay, just let me know. I appreciate any volunteer effort I can get, so I won't complain :D

cmd/mockery.go
LandonTClipp: I think we should leave it as-is. It's kind of poorly named, but I will make a note of it in the docs clearly explaining what it's used for. In v3 we can probably change the name to make it more obvious.

mcdoker18:
@LandonTClipp I noticed couple of minor issues:
* we can't use the option `tags` in the package configs. Because mockery takes only top level `tags` and passes it to he package loader.
* Now the `tags` option is space-separated list, but in all other tools, like go build or go test, it is comma-separated list.

cmd/mockery.go
mcdoker18: I added the new option and changed derscription of the `tags` option.

mocks/github.com/vektra/mockery/v2/pkg/fixtures/buildtag/comment/IfaceWithBuildTagInComment.go
mcdoker18: Fixed

mcdoker18: By the way [boilerplate text](https://github.com/vektra/mockery/blob/138d507e2b506420f49969570b816f31287b8bed/pkg/generator.go#L112C4-L112C23) is generated before `// Code generated ...`. Is it correct?

docs/configuration.md
```diff
@@ -78,7 +78,8 @@ Parameter Descriptions
...
-| `tags`                                                 | :fontawesome-solid-x:     | `#!yaml ""`                           | Set the build tags of the generated mocks.                                                                                                                                                                                                           |
+| `tags`                                                 | :fontawesome-solid-x:     | `#!yaml ""`                           | A space-separated list of additional build tags to load packages.                                                                                                                                                                                                                |
+| `mock-build-tags`                                       | :fontawesome-solid-x:     | `#!yaml ""`                          | Set the build tags of the generated mocks. Read more about the [format](https://pkg.go.dev/cmd/go#hdr-Build_constraints).                                                                                                                                                        |
```
LandonTClipp, line 82: Can you place this alphabetically?

.mockery.yaml
```diff
@@ -4,7 +4,14 @@ with-expecter: True
 mockname: "{{.InterfaceName}}"
 filename: "{{.MockName}}.go"
 outpkg: mocks
+tags: "custom2"
```
LandonTClipp, line 7: What's the purpose of adding this?

LandonTClipp (changes requested):
This is 99% of the way there, just a couple questions. Thanks again for helping out, I really do appreciate it.

docs/configuration.md
mcdoker18: Done.

.mockery.yaml
mcdoker18: I check that passing of build tags to package loader works in e2e tests. The IfaceWithCustomBuildTagInComment mock will not be generated without this line.

You can check by commenting this line and exec `task test.e2e`.

LandonTClipp: Ah got it! Makes sense.

LandonTClipp (approved):
Looks great, approved. I'll merge this into master and create a new build.

Since this is _technically_ adding new functionality, I'll bump it as a minor revision.

PR #695, opened 2023-08-18
Allow setting ExpectedCalls on mock constructor

### LINKED ISSUE #696: Construct and bootstrap mock more concisely
Description
-------------
Hi there 👋🏼

We use table driven tests with mockery on a day to day basis. These tests generally look something like this:
```golang
func TestUserExists(t *testing.T) {
	var userID int64 = 12345
	tcs := []struct {
		name      string
		userMocks []*mock.Call
		expExists bool
		expErr    error
	}{
		{
			name: "user exists",
			userMocks: []*mock.Call{
				new(mock.Mock).On("Lookup", mock.Anything, userID).Return(&users.User{
					Id: userID,
				}, nil),
			},
			expExists: true,
		},
		{
			name: "user does not exists",
			userMocks: []*mock.Call{
				new(mock.Mock).On("Lookup", mock.Anything, userID).Return(nil, errUserNotFound),
			},
			expExists: false,
		},
		{
			name: "user lookup failure",
			userMocks: []*mock.Call{
				new(mock.Mock).On("Lookup", mock.Anything, userID).Return(nil, context.Canceled),
			},
			expExists: false,
			expErr:    context.Canceled,
		},
	}

	for _, tt := range tcs {
		t.Run(tt.name, func(t *testing.T) {
			ctx := context.Background()
			uc := users_mock.NewClient(t) // Here's the mock instantiation
			uc.ExpectedCalls = tt.userMocks

			exists, err := UserExists(ctx, uc, userID)
			require.ErrorIs(t, err, tt.expErr)
			require.Equal(t, tt.expExists, exists)
		})
	}
}
```

The above test doesn't look too bad but the multiple lines needed to instantiate a mock client and attach the expected calls can impact the function's readability when there are many different kinds of mocks in the test.

In most cases we also use a dependency injection pattern where a bunch of dependencies (in the form of mocks) need to be instantiated. This looks something like this:
```golang
uc := users_mock.NewClient(t)
uc.ExpectedCalls = tt.userMocks
deps := Dependencies{
	UsersClient: uc,
}
// Use deps in tested function
```
The list of dependencies can grow fairly large depending on the complexity of the function being tested. 

The change I'm proposing in https://github.com/vektra/mockery/pull/695 allows a mock to be instantiated and bootstrapped on one line. This enables the above two examples to be much easier to read and the setup of a unit test to be more concise.

I'm sure there are other ways to achieve the same goal. I'm keen to get your feedback on the proposed change, or any alternative that you may suggest.

Mockery Version
--------

v2@latest

Golang Version
--------------
Latest

Vorster4427:
Description
-------------

The proposed change makes the following possible on the test in the linked issue:
```golang
uc := users_mock.NewClient(t, tt.userMocks...)
``` 

This makes the test more readable and makes the mockery implementation of an interface easier to use in table driven tests.

This is my first attempt at an open source contribution, so I'm happy to update any code styles or follow a maintainer's lead on how to best implement this change.

Closes #696 

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

I used the new constructor signature to write the unit test above (the test passed). I've also updated the `mocks/` directory to include the new constructor signature in the generated code test files.

Checklist
-----------

### DIFF
--- a/mocks/github.com/vektra/mockery/v2/pkg/TypesPackage.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/TypesPackage.go
@@ -104,9 +104,10 @@ func (_c *TypesPackage_Path_Call) RunAndReturn(run func() string) *TypesPackage_
 func NewTypesPackage(t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *TypesPackage {
+}, expectedCalls ...*mock.Call) *TypesPackage {
 	mock := &TypesPackage{}
 	mock.Mock.Test(t)
+	mock.Mock.ExpectedCalls = expectedCalls
 
 	t.Cleanup(func() { mock.AssertExpectations(t) })
 
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/A.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/A.go
@@ -76,9 +76,10 @@ func (_c *A_Call_Call) RunAndReturn(run func() (test.B, error)) *A_Call_Call {
 func NewA(t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *A {
+}, expectedCalls ...*mock.Call) *A {
 	mock := &A{}
 	mock.Mock.Test(t)
+	mock.Mock.ExpectedCalls = expectedCalls
 
 	t.Cleanup(func() { mock.AssertExpectations(t) })
 
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/AsyncProducer.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/AsyncProducer.go
@@ -151,9 +151,10 @@ func (_c *AsyncProducer_Whatever_Call) RunAndReturn(run func() chan bool) *Async
 func NewAsyncProducer(t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *AsyncProducer {
+}, expectedCalls ...*mock.Call) *AsyncProducer {
 	mock := &AsyncProducer{}
 	mock.Mock.Test(t)
+	mock.Mock.ExpectedCalls = expectedCalls
 
 	t.Cleanup(func() { mock.AssertExpectations(t) })
 
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Blank.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Blank.go
@@ -64,9 +64,10 @@ func (_c *Blank_Create_Call) RunAndReturn(run func(interface{}) error) *Blank_Cr
 func NewBlank(t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *Blank {
+}, expectedCalls ...*mock.Call) *Blank {
 	mock := &Blank{}
 	mock.Mock.Test(t)
+	mock.Mock.ExpectedCalls = expectedCalls
 
 	t.Cleanup(func() { mock.AssertExpectations(t) })
 
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/ConsulLock.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/ConsulLock.go
@@ -117,9 +117,10 @@ func (_c *ConsulLock_Unlock_Call) RunAndReturn(run func() error) *ConsulLock_Unl
 func NewConsulLock(t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *ConsulLock {
+}, expectedCalls ...*mock.Call) *ConsulLock {
 	mock := &ConsulLock{}
 	mock.Mock.Test(t)
+	mock.Mock.ExpectedCalls = expectedCalls
 
 	t.Cleanup(func() { mock.AssertExpectations(t) })
 
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/EmbeddedGet.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/EmbeddedGet.go
@@ -66,9 +66,10 @@ func (_c *EmbeddedGet_Get_Call[T]) RunAndReturn(run func() T) *EmbeddedGet_Get_C
 func NewEmbeddedGet[T constraints.Signed](t interface {
 	mock.TestingT
 	Cleanup(func())
-}) *EmbeddedGet[T] {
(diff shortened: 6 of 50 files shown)


LandonTClipp:
Hey thanks for the PR. I'm not sure I want to implement this since the "approved" way of doing this is simply modifying the mock attribute like you mentioned. Proliferating these arguments makes no sense if there's already an easy way to do it.

echarrod:
> Hey thanks for the PR. I'm not sure I want to implement this since the "approved" way of doing this is simply modifying the mock attribute like you mentioned. Proliferating these arguments makes no sense if there's already an easy way to do it.

I agree that modifying the mock attribute is simple, **but** if you want to instantiate mock clients in a single line, so that you're able to declare many in each table-driven test, there's no way of doing that currently I believe.

This would allow for that, and makes the tests able to be written much more succinctly, unless I'm missing something, and there's a better way?

LandonTClipp:
> but if you want to instantiate mock clients in a single line, so that you're able to declare many in each table-driven test, there's no way of doing that currently I believe.

I'm not sure I understand this bit. Are you saying that you would rather declare the whole mock object (including expectations) inside of the table elements itself, or just that you would rather have the mock declaration from within the for loop be only one line?

Vorster4427:
> I'm not sure I understand this bit. Are you saying that you would rather declare the whole mock object (including expectations) inside of the table elements itself, or just that you would rather have the mock declaration from within the for loop be only one line?

Thanks for the reply, Landon. The goal is the latter in your statement. To be able to instantiate the mock object within the for loop in one line.  So the table elements contain the expectations and the mock object is instantiated within the `t.Run` callback on one line.

LandonTClipp:
Okay thanks for your clarification, I understand. I really appreciate your enthusiasm for improving the project, however I'm of the opinion that this feature request doesn't have enough justification for the following reasons:

1. Modifying the ExpectedCalls attribute is fairly straightforward.
2. Saving only one line of code isn't that huge of a win
3. Adding arguments to the constructor needs to be done in a type-agnostic way (meaning, we should use [functional options instead](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). This PR could be easily modified to do that, so this reason isn't necessarily a point against the feature itself, but rather just the proposed implementation.
4. The most relevant point I think: creating expectations in this way is technically not the preferred method. Instead, using `.EXPECT()` is preferred due to the amount of type safety it gives you. Using bare `.On(` calls isn't type safe. So, I tend to not encourage it, and adding new features that would encourage its use is antithetical to the direction I'm trying to take.

As always, I guide this project in the direction that people generally want it to go (or the direction I think they want it to go). So if something like this becomes a huge deal for a lot of people then I can consider it more strongly, but otherwise I think for now it's going to be a no-go. But, thanks for interacting anyway and keep the ideas coming, that's what makes this project great!

Vorster4427:
Thank you for the detailed explanation. We recently started using the expecter feature and the type safety it provides is great. I'm happy to abandon the proposed change given your points above. 

I would like to get your opinion on what a table driven test _should_ look like when using `.EXPECT()`. 

Given the same test from the issue but now with `.EXPECT()`
```golang
func TestUserExists(t *testing.T) {
	var userID int64 = 12345
	tcs := []struct {
		name      string
		userMocks []*mock.Call
		expExists bool
		expErr    error
	}{
		{
			name: "user exists",
			userMocks: []*mock.Call{
				new(users_mock.Client).EXPECT().Lookup(mock.Anything, userID).Return(&users.User{
					Id: userID,
				}, nil).Once(),
			},
			expExists: true,
		},
		{
			name: "user does not exists",
			userMocks: []*mock.Call{
				new(users_mock.Client).EXPECT().Lookup(mock.Anything, userID).Return(nil, errUserNotFound).Once(),
			},
			expExists: false,
		},
		{
			name: "user lookup failure",
			userMocks: []*mock.Call{
				new(users_mock.Client).EXPECT().Lookup(mock.Anything, userID).Return(nil, context.Canceled).Once(),
			},
			expExists: false,
			expErr:    context.Canceled,
		},
	}

	for _, tt := range tcs {
		t.Run(tt.name, func(t *testing.T) {
			ctx := context.Background()
			uc := users_mock.NewClient(t) // Here's the mock instantiation
			uc.ExpectedCalls = tt.userMocks

			exists, err := UserExists(ctx, uc, userID)
			require.ErrorIs(t, err, tt.expErr)
			require.Equal(t, tt.expExists, exists)
		})
	}
}
```

Would you change anything in the structure of the test? 

The above is currently what our tests look like when using `.EXPECT()`, but it feels awkward instantiating a new `Client` for each expectation. It could be that we're just not using `mockery` the way it's intended, so we're trying to find solutions to problems we create ourselves.

LandonTClipp:
I don't think it seems all that awkward. What I've done is something like this:

```go
mockClient: func(t *testing.T) *MockClient {
        NewMockClient(m).EXPECT().Lookup(mock.Anything, userID).Return(nil, context.Canceled).Once()
        return m
},
```

then in the test body

```go
	for _, tt := range tcs {
		t.Run(tt.name, func(t *testing.T) {
                        client := tt.mockClient(t)
                        FooBar(client)
		})
	}

```

Vorster4427:
That's an interesting approach. Thanks for sharing!

I'll close the issue and the PR.

echarrod:
Thanks for the response @LandonTClipp.
I just tried implementing this method with you shared, and it certainly is quite a bit clunkier!

You said: 
> It's not awkward, this is actually how the object is supposed to be used. You should create a totally new object each time your test runs

But because we're instantiating the mock client inside a `t.Run(tt.name, func(t *testing.T) {`, a totally new object _is_ created each time the test runs, right?

It seems to add a lot of noise, and make the tests harder to read, in my opinion, if we have to have this inside _every_ test case:
```go
mockClient: func(t *testing.T) *MockClient {
        m := NewMockClient(t)
        m.EXPECT().Lookup(mock.Anything, userID).Return(nil, context.Canceled).Once()
        return m
},
```
Rather than this
```go
userMocks: []*mock.Call{
	new(users_mock.Client).EXPECT().Lookup(mock.Anything, userID).Return(nil, context.Canceled).Once(),
},
```

echarrod:
But maybe this is a side-track, as looking back, I believe the goal of this change was to be able to combine the following two lines:
```go
uc := users_mock.NewClient(t)
uc.ExpectedCalls = tt.userMocks
```
Into:
```go
uc := users_mock.NewClient(t, tt.userMocks...)
```

PR #699, opened 2023-08-30
fix expecter and void rolled varaidic

jokly:
Description
-------------

If
```yaml
with-expecter: True
unroll-variadic: False
```
and varaidic function doesn't return arguments then the wrong mock is generated. Example:
```golang
// --- interface ---
type ILogger interface {
    Infof(format string, a ...interface{})
}

// --- mock ---
func (_m *MockIActionContext) Infof(format string, a ...interface{}) {
	var tmpRet mock.Arguments
	if len(a) > 0 {
		tmpRet = _m.Called(format, a)
	} else {
		tmpRet = _m.Called(format)
	}

	tmpRet // <--- error:  tmpRet (variable of type mock.Arguments) is not used
}
```
- Fixes #629

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

Add new test case
```golang
VariadicNoReturn(j int, is ...interface{})
```

Checklist
-----------


pkg/generator.go
```diff
@@ -930,10 +931,20 @@ func (g *Generator) generateCalled(list *paramList) (preamble string, called str
...
-			preamble = fmt.Sprintf("\n\tvar " + tmpRet + " mock.Arguments\n\tif len(" + variadicName + ") > 0 {\n\t\t" + tmpRet + " = _m.Called(" + strings.Join(list.Names, ", ") + ")\n\t} else {\n\t\t" + tmpRet + " = _m.Called(" + strings.Join(list.Names[:len(list.Names)-1], ", ") + ")\n\t}\n\n\t")
-			called = tmpRet
+			called = fmt.Sprintf(
+				`func() mock.Arguments {
+	if len(%s) > 0 {
+		return _m.Called(%s)
+	} else {
+		return _m.Called(%s)
+	}
+}()`,
```
LandonTClipp, line 942: This is a much better way to format the string, thanks for the improvement.

I understand what this is doing and why you made this particular anonymous function, but it _feels_ clunky to me, and non-obvious. Ideally the generated code would be something like:

```go
if len(%s) > 0 {
    _m.Called(%s)
} else {
    _m.Called(%s)
}
```

when the variadic has no return and

```go
var tmpRet mock.Arguments
if len(ints) > 0 {
    tmpRet = _m.Called(ints)
} else {
    tmpRet = _m.Called()
}

ret := tmpRet
```

when there is a return. 

Your implementation is basically tricking the compiler into allowing us to not use the return value of `_m.Called` so in that sense I get it, but it makes the mock code look weird.

I think instead what I'd like to see here is something like

```go
// This function would be something new you have to make.
func (g *Generator) printTemplateBytes(data interface{}, templateString string) *bytes.Buffer {
	var buf bytes.Buffer
	tmpl, err := templates.New(templateString).Funcs(templateFuncMap).Parse(templateString)
	if err != nil {
		// couldn't compile template
		panic(err)
	}

	if err := tmpl.Execute(&buf, data); err != nil {
		panic(err)
	}
	return &buf
}

// Modify the existing `printTemplate` function to instead call out to `printTemplateBytes`
func (g *Generator) printTemplate(data interface{}, templateString string) {
	g.buf.Write(g.printTemplateBytes(data, templateString).Bytes())
}

// Now because we can write the template directly, it becomes a bit easier to do what I'm
// suggesting
func (g *Generator) generateCalled(list *paramList) (preamble string, called string) {
	namesLen := len(list.Names)
	if namesLen == 0 || !list.Variadic || !g.config.UnrollVariadic {
		if list.Variadic && !g.config.UnrollVariadic && g.config.WithExpecter {

			var tmpRet, tmpRetWithAssignment string
			if namesLen != 0 {
				tmpRet = "tmpRet"
				tmpRetWithAssignment = "tmpRet = "
			}
			calledBytes := g.printTemplateBytes(
				struct {
					paramList                 *paramList
					paramNamesWithoutVariadic []string
					tmpRet                    string
					tmpRetWithAssignment      string
					variadicName              string
				}{
					paramList:                 list,
					paramNamesWithoutVariadic: list.Names[:len(list.Names)-1],
					tmpRet:                    tmpRet,
					tmpRetWithAssignment:      tmpRetWithAssignment,
					variadicName:              list.Names[namesLen-1],
				},
				`
	{{ .tmpRet }}
	if len({{ .variadicName }) > 0 {
		{{ .tmpRetWithAssignment }}_m.Called({{ join .paramList.Names ", " }})
	} else {
		{{ .tmpRetWithAssignment }}_m.Called({{ paramNamesWithoutVariadic  }})
	}
	`,
			)

			return tmpRet, calledBytes.String()
		}

		called = "_m.Called(" + strings.Join(list.Names, ", ") + ")"
		return
	}
...
```

What this gives us is mock code that doesn't rely on this anonymous function hack, it properly removes the `tmpRet` assignments if there are no variables to return.

pkg/fixtures/expecter.go
```diff
@@ -6,4 +6,5 @@ type Expecter interface {
 	ManyArgsReturns(str string, i int) (strs []string, err error)
 	Variadic(ints ...int) error
 	VariadicMany(i int, a string, intfs ...interface{}) error
+	VariadicNoReturn(j int, is ...interface{})
```
LandonTClipp, line 9: I recommend creating a new interface fixture and then in `.mockery.yaml` add a new configuration section using the previously buggy config. This is going to assert that mockery generates valid code. Something like

```yaml
      VariadicNoReturnInterface:
        config:
          with-expecter: True
          unroll-variadic: False
```

This way the integration tests will run and assert the validity of the generated mocks.

pkg/generator.go
LandonTClipp: In fact, I ran the code above and it creates the mock as expected

```go
// VariadicNoReturn provides a mock function with given fields: j, is
func (_m *VariadicNoReturnInterface) VariadicNoReturn(j int, is ...interface{}) {

	if len(is) > 0 {
		_m.Called(j, is)
	} else {
		_m.Called(j)
	}

}
```

I basically wrote my whole implementation already 😅 but I'll still let you copy-paste-modify-improve however you want since you gave it a good first stab.

LandonTClipp (changes requested):
I have a number of changes I commented below regarding tests and general strategy for this bug.

jokly:
Thank you for the review. I've made changes as you recommend :)

LandonTClipp (approved):

LandonTClipp:
Looks great! Thank you very much for the fix.

Issue #708, opened 2023-09-21
Support "exclude" in packages config


kpaulisse:
Description
-------------

The "exclude" parameter is not recognized when the packages configuration is used. Including this parameter in the configuration has no effect.

Mockery Version
--------

v2.33.2 and latest `master`

Golang Version
--------------

1.20.8

Installation Method
-------------------

- [x] Binary Distribution
- [ ] Docker
- [x] brew
- [ ] go install
- [x] Other: Compiling from source

Steps to Reproduce
------------------

We have a package in our code base that doesn't have any interfaces but consistently fails with errors like this:

```
21 Sep 23 15:03 CDT ERR encountered error when loading package error="/Users/xxxxxxxx/go/src/github.com/organization/reponame/blah/blah/clients/common/common_test_utils.go:4:10: could not import github.com/organization/reponame/blah/blah/mocks/clients/foo (invalid package name: \"\")" dry-run=false version=v0.0.0-dev
```

The `common_test_utils.go` file featured in that error message a test helper that contains code referencing mocks like this:

```go
package clients

import (
	mockFoo "github.com/organization/reponame/blah/blah/mocks/clients/foo"
)
```

I want to use this configuration in my `.mockery.yaml` file, but there's no code in `config.go` that recognizes the values in the "Exclude" field:

```yaml
---
with-expecter: False
filename: "{{.InterfaceName}}.go"
mockname: "{{.InterfaceName}}"
outpkg: "mocks"
packages:
  github.com/organization/reponame/blah/blah/clients:
    config:
      all: true
      dir: mocks/{{ .InterfaceDirRelative }}
      recursive: true
      exclude:
        - common
```

### Expected Behavior

It should not descend into the "common" directory and therefore I should not get an error when it fails to process the file in that directory correctly.

### Actual Behavior

It descends into the "common" directory and I get the error shown above.

tzachshabtay:
@LandonTClipp @kpaulisse is this part of mockery 2.36.0?

My "exclude" direction in the config file is ignored, not sure if I'm doing it wrong?

i.e this is my config file:

```
with-expecter: true
dir: "{{.InterfaceDir}}/mock" 
outpkg: "mock"
filename: "{{ .InterfaceName | snakecase }}.go"
mockname: "{{ .InterfaceName | firstUpper }}"
packages:
  org/bla/bla/package1:
    config:
      recursive: True
      exclude:
        - reinforced
      include-regex: ".*"
[...]

PR #709, opened 2023-09-21
Support "exclude" in package config

kpaulisse:
Description
-------------

This PR recognizes the "exclude" directive in a package configuration. The following configuration snippet now excludes the "client/common" directory (and everything therein):

```yaml
---
with-expecter: False
filename: "{{.InterfaceName}}.go"
include-auto-generated: false
mockname: "{{.InterfaceName}}"
outpkg: "mocks"
packages:
  github.com/organization/reponame/core/clients:
    config:
      all: true
      dir: mocks/{{ .InterfaceDirRelative }}
      recursive: true
      exclude:
        - common
```

This is motivated by errors we get from mockery arising from imports in `github.com/organization/reponame/core/clients/common`. We want to be able to use the recursive functionality (since we have a bunch of different clients) but skip the one directory that has only test helpers and no interfaces to mock. The "exclude" feature lets us do that in the old style configuration format, but there was no implementation of this in the new packages format.

Fixes https://github.com/vektra/mockery/issues/708

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

Added a unit test to the existing table driven test that exercises the revised functionality. (It may look a bit strange to add a stanza where the input YAML equals the output YAML. However without the change in the code, the output YAML would be auto-populated with the information about `subpkg3`).

Tested in our use case, as running `mockery` against our configuration (represented by the YAML snippet above) correctly logs that it is skipping the subpackage and exits without error.

Checklist
-----------

Regarding documentation updates: `exclude` is in the table of deprecated parameters for v3. I am hoping that this use case allows it to be brought back.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -599,7 +599,7 @@ func (c *Config) discoverRecursivePackages(ctx context.Context) error {
 				p := pathlib.NewPath(subPkg)
 				relativePath, err := p.RelativeToStr(pkgPath)
 				if err != nil {
-					return fmt.Errorf("failed to get path for %s relative to %s: %w", subPkg, pkgPath, err)
+					return stackerr.NewStackErrf(err, "failed to get path for %s relative to %s", subPkg, pkgPath)
 				}
 				if conf.ExcludePath(relativePath.String()) {
 					subPkgLog.Info().Msg("subpackage is excluded")
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -595,6 +595,18 @@ func (c *Config) discoverRecursivePackages(ctx context.Context) error {
 			subPkgLog := pkgLog.With().Str("sub-package", subPkg).Logger()
 			subPkgCtx := subPkgLog.WithContext(pkgCtx)
 
+			if len(conf.Exclude) > 0 {
+				p := pathlib.NewPath(subPkg)
+				relativePath, err := p.RelativeToStr(pkgPath)
+				if err != nil {
+					return fmt.Errorf("failed to get path for %s relative to %s: %w", subPkg, pkgPath, err)
+				}
+				if conf.ExcludePath(relativePath.String()) {
+					subPkgLog.Info().Msg("subpackage is excluded")
+					continue
+				}
+			}
+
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -595,6 +595,18 @@ func (c *Config) discoverRecursivePackages(ctx context.Context) error {
 			subPkgLog := pkgLog.With().Str("sub-package", subPkg).Logger()
 			subPkgCtx := subPkgLog.WithContext(pkgCtx)
 
+			if len(conf.Exclude) > 0 {
+				p := pathlib.NewPath(subPkg)
+				relativePath, err := p.RelativeToStr(pkgPath)
+				if err != nil {
+					return fmt.Errorf("failed to get path for %s relative to %s: %w", subPkg, pkgPath, err)
```
LandonTClipp, line 602: Please change this to use the StackErr: https://github.com/vektra/mockery/blob/master/pkg/stackerr/stackerr.go#L21

pkg/config/config_test.go
```diff
@@ -980,6 +980,33 @@ packages:
...
+packages:
+  github.com/vektra/mockery/v2/pkg/fixtures/example_project/pkg_with_subpkgs/subpkg2:
+    config:
+      recursive: True
+      with-expecter: True
+      all: True
+      exclude:
+        - subpkg3
+`,
+			wantCfgMap: `dir: foobar
```
LandonTClipp, line 999: I understand about the funkiness of the test. That's okay, people can come back to this PR if they're confused :)

LandonTClipp (changes requested):
I have one change but otherwise this is good to go. Please address that and we can get this merged.

pkg/config/config.go
kpaulisse: Done in 1407ff88ea428c24756ba122d01fc79392977b7e

LandonTClipp (approved):
Looks good, will merge when tests pass.

PR #710, opened 2023-09-23
Fix replace-type for different packages from the same source

### LINKED ISSUE #688: --replace-type is not mapping packages correctly
Description
-------------

I am trying to mock an interface that requires multiple ```--replace-type``` inputs. The target type aliases for these types reside in separate packages but the source internal types are all in the same package. This causes the generate logic to choose the wrong alias when building the mock.

The issue seems to be caused by ```addPackageImportWithName``` only looking at the the package portion of the from ```replaceType``` [here](https://github.com/vektra/mockery/blob/master/pkg/generator.go#L191). It should probably scan all the replace type inputs first to see if there is an explicit type-level mapping and then a follow up scan for package-level mappings.

Mockery Version
--------

2.30.1, 2.32.3

Golang Version
--------------

1.20

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [x] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

I set up a reproducer repo [here](https://github.com/rkoehl05/mockrepro). You can run go generate. in it to produce the broken mock.

### Expected Behavior

The ```DeleteResponse``` type should use the ```blob``` alias, not ```blockblob```

### Actual Behavior

```DeleteResponse``` is incorrectly referenced using the ```blockblob``` alias. See [here](https://github.com/rkoehl05/mockrepro/blob/main/example_output.txt#L82) for an example.

### PR DESCRIPTION
**Pull request by RangelReale**
Description
-------------

When different packages that map to the same source package and mapped with replace-type, the package of the first one are used for all other packages.

- Fixes #688

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [ ] 1.20
- [x] 1.21

How Has This Been Tested?
---------------------------

- a test was added for this specific problem.
- @rkoehl05 was able to validate the solution in his mock repo.

Checklist
-----------

### DIFF
--- a/pkg/generator.go
+++ b/pkg/generator.go
@@ -103,7 +103,7 @@ func NewGenerator(ctx context.Context, c GeneratorConfig, iface *Interface, pkg 
 	}
 
 	g.parseReplaceTypes(ctx)
-	g.addPackageImportWithName(ctx, "github.com/stretchr/testify/mock", "mock")
+	g.addPackageImportWithName(ctx, "github.com/stretchr/testify/mock", "mock", nil)
 
 	return g
 }
@@ -161,7 +161,7 @@ func (g *Generator) getPackageScopedType(ctx context.Context, o *types.TypeName)
 		(!g.config.KeepTree && g.config.InPackage && o.Pkg() == g.iface.Pkg) {
 		return o.Name()
 	}
-	pkg := g.addPackageImport(ctx, o.Pkg())
+	pkg := g.addPackageImport(ctx, o.Pkg(), o)
 	name := o.Name()
 	g.checkReplaceType(ctx, func(from *replaceType, to *replaceType) bool {
 		if o.Pkg().Path() == from.pkg && name == from.typ {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by RangelReale** [comment]
@LandonTClipp had to add the type to the `addPackageImportWithName` to be able to differentiate between them, I only replace if this parameter is not nil. Also added some specifity rule to check replaces with types first.

@rkoehl05 can you see if this fixes your problem?

**Comment by rkoehl05** [comment]
Thanks @RangelReale! I tested with my reproducer and I was able to generate a working mock.

**Comment by LandonTClipp** [comment]
@RangelReale thanks so much for this. I converted this to a draft because you mentioned this is still a WIP. Please ping me when you're ready for me to review.

**Comment by RangelReale** [comment]
> @RangelReale thanks so much for this. I converted this to a draft because you mentioned this is still a WIP. Please ping me when you're ready for me to review.

@LandonTClipp I updated the description, it should be ready now.

File: pkg/fixtures/example_project/fiz/fiz.go [hunk 1]
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/vektra/mockery/v2/pkg/fixtures/example_project/fiz/fizi/fiz1"
+	"github.com/vektra/mockery/v2/pkg/fixtures/example_project/fiz/fizi/fiz2"
+)
+
+type Fiz interface {
+	FA(f fiz1.Fiz1)
+	FB(f fiz2.Fiz2)
+}
```
LandonTClipp, line 11: I would request you add a short README.md to `example_project/fiz` explaining what this package is testing.

Also I'd like it to be renamed to something more descriptive like `replaceType` or something.

pkg/generator.go
```diff
@@ -173,23 +173,28 @@ func (g *Generator) getPackageScopedType(ctx context.Context, o *types.TypeName)
...
+func (g *Generator) addPackageImport(ctx context.Context, pkg *types.Package, o *types.TypeName) string {
+	return g.addPackageImportWithName(ctx, pkg.Path(), pkg.Name(), o)
 }
 
 func (g *Generator) checkReplaceType(ctx context.Context, f func(from *replaceType, to *replaceType) bool) {
-	for _, item := range g.replaceTypeCache {
-		if !f(item.from, item.to) {
-			break
+	// check most specific first
+	for _, hasType := range []bool{true, false} {
```
LandonTClipp, line 182: Good solution 👍🏻

LandonTClipp (changes requested):
Good implementation. Just minor nitpicks about the naming that could make this a bit clearer.

pkg/fixtures/example_project/fiz/fiz.go
RangelReale: renamed the files and types, and added a README.

LandonTClipp (approved):
Awesome thank you!

LandonTClipp:
@RangelReale if you could squash the commits down to one, we can merge away.

RangelReale:
> @RangelReale if you could squash the commits down to one, we can merge away.

rebased to master, see if this is ok.

LandonTClipp:
Thank you very much for getting this out. I lean heavily on people submitting PRs ❤️

PR #736, opened 2023-11-20
Adds useful message when test panics on no return values

### LINKED ISSUE #729: Omitting `.Return()` on an expectation results in a panic and a stack trace rather than a useful error message
Description
-------------

Omitting the `.Return()` on an expectation results in a panic deep in a transitive dependency that does not indicate what the problem was or what the resolution might be.

Mockery Version
--------

v2.36.0

Golang Version
--------------

go version go1.21.3 darwin/arm64

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [x] brew
- [ ] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

When setting an expectation on a function call defined thus:
```
DoThing(operatorID string, classTypeID *string, input []TextInput) error
```
It will look something like:
```
serviceAPI.EXPECT().DoThing("operator-id", (*string)(nil), []serviceAPI.TextInput{
    {ReportID: reportID.String(), Text: "report-text"},
})
```
(note missing `.Return(nil)` at the end of the `EXPECT()`, an easy mistake to make.)

### Expected Behavior

In this situation I would expect an error message saying something like "Required `.Return(error)` not specified on `EXPECT()` call for `DoThing()` invocation on line xxx".

### Actual Behavior

I get a panic and a stack trace from a transitive dependency (not mockery) that gives me no idea what the error could possibly be or how to fix it:
```
  assert: arguments: Cannot call Get(0) because there are 0 argument(s).

  Full Stack Trace
    github.com/stretchr/testify/mock.Arguments.Get(...)
    	/Users/ratkins/go/pkg/mod/github.com/stretchr/testify@v1.8.4/mock/mock.go:887
    gitlab.com/xxx/xxx/mocks/gitlab.com/xxx/xxx/src/services/xxx.(*MockAPI).DoThing(0x140000328c0, {0x103229442, 0xb}, 0x0, {0x140003ec700, 0x1, 0x1})
    	/Users/ratkins/Developer/xxx/mocks/gitlab.com/xxx/xxx/src/services/xxx/mock_API.go:28 +0x1c8
    gitlab.com/xxx/xxx/src/endpoints/reports_test.glob..func1.1.MakePutHandler.func1(0x140000328c0?)
    	/Users/ratkins/Developer/xxx/src/endpoints/reports/reports_put.go:25 +0x1b8
    gitlab.com/xxx/xxx/src/endpoints/reports_test.glob..func1.1()
```
---
I realise this bug is possibly testify's "fault", but I didn't install testify, I installed mockery, and not being deeply into the Go ecosystem it took me a while to work out what the relationship is.

mateusmarquezini:
Description
-------------
- Adds a useful message when the test panics on no return value.

- Fixes #729  

## Type of change

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [x] 1.20

How Has This Been Tested?
---------------------------

I tested by using Unit tests

Checklist
-----------


pkg/generator_test.go
```diff
@@ -517,6 +517,10 @@ func (s *GeneratorSuite) TestGeneratorForEmptyInterface() {
 	s.checkGeneration("empty_interface.go", "Blank", false, "", "")
 }
 
+func (s *GeneratorSuite) TestGeneratorForPanicOnNoReturnValueInterface() {
+	s.checkGeneration("panic_err.go", "PanicOnNoReturnValue", false, "", "")
+}
+
```
LandonTClipp, line 523: This is not a useful style of test, I'm trying to move away from these. Just create a test that asserts the behavior of the mock object.

Thanks for tackling this!

mateusmarquezini: Hey @LandonTClipp! Thanks for suggesting this! 

Yeah, I just created the test for the new message when the test panics and deleted this one. Thanks for the heads up! I thought this was something necessary as well.

pkg/fixtures/test/panic_on_no_return_value_test.go
```diff
@@ -0,0 +1,23 @@
+package test
```
mateusmarquezini, line 1: @LandonTClipp I was wondering if this is the right package for this test, but as you suggested to create a new one, I created a new file and placed it down next to the fixture that I just created, Let me know if there is another better package for this.

LandonTClipp: You can just put it in the same directory as the fixture. Thanks!

pkg/fixtures/test/panic_on_no_return_value_test.go
```diff
@@ -0,0 +1,23 @@
...
+	mocks "github.com/vektra/mockery/v2/mocks/github.com/vektra/mockery/v2/pkg/fixtures"
+)
+
+func TestPanicOnNoReturnValue(t *testing.T) {
+	m := mocks.NewPanicOnNoReturnValue(t)
+	m.EXPECT().DoSomething()
+
+	defer func() {
+		if r := recover(); r != nil {
+			assert.NotNil(t, r)
```
LandonTClipp, line 16: This needs to be `require.NotNil` so that the test will fail immediately. Otherwise the `r.(string)` cast would panic.

pkg/fixtures/test/panic_on_no_return_value_test.go
```diff
@@ -0,0 +1,23 @@
...
+			assert.Equal(t, "Missing Return() function for DoSomething()", r.(string))
+		}
+	}()
+
+	m.DoSomething()
+
```
LandonTClipp, line 22: You'll need some sort of assertion that the panic actually happened. Since the recovery is happening within a defer, you might need another deferred statement that asserts the panic occurred. Maybe

Defers are in LIFO order so maybe something like

```go
var panicOccurred bool
defer func() {
    assert.True(t, panicOccurred)
}()
defer func() {
    panicOccurred = true
    // additional recovery logic you already have
}

m.DoSomething()
```

pkg/generator.go
```diff
@@ -752,6 +752,10 @@ func (_m *{{.MockName}}{{.InstantiatedTypeString}}) {{.FunctionName}}({{join .Pa
 {{- else}}
 	{{- .RetVariableName}} := {{.Called}}
 
+	if len({{.RetVariableName}}) == 0 {
+		panic("Missing Return() function for {{.FunctionName}}()")
```
LandonTClipp, line 756: Alternatively, I'd say the message should be:

```
"no return value specified for {{.FunctionName}}"
```

Return values can be specified in multiple different ways, not just through `.Return()`.

pkg/fixtures/test/panic_on_no_return_value_test.go
mateusmarquezini: Thanks for all your code suggestions, @LandonTClipp! At this specific point, I guess it won't be possible as we have an import cycle here because of one of the imports used in the test file [here](https://github.com/vektra/mockery/pull/736/files#diff-eaff0e9a7c5657f64f7817700c73dd7ed611083fd46e32faced795b83ffa7873R8).

LandonTClipp: Not sure I understand. Placing the test next to `pkg/fixtures/panic_err.go` makes an import cycle?

pkg/fixtures/test/panic_on_no_return_value_test.go
```diff
@@ -0,0 +1,29 @@
...
+	m.EXPECT().DoSomething()
+
+	var panicOccurred bool
+	defer func() {
+		assert.True(t, panicOccurred)
+	}()
+	defer func() {
+		panicOccurred = true
+		if r := recover(); r != nil {
+			require.NotNil(t, r)
```
LandonTClipp, line 22: Actually one more thing about this. Don't use the if statement, it's possible in this test that we're silently not asserting the contents of `r` (in the case it's nil.

```go
r := recover()
require.NotNil(t, r)
assert....
```

mateusmarquezini: Great, thanks!

pkg/fixtures/test/panic_on_no_return_value_test.go
mateusmarquezini: I discovered why we had an import cycle here, the package from the **pkg/fixtures/panic_err.go**  file was named "`package test`" (like in the other files next to it), and the package in **panic_err_test.go** was also named "test", so renaming it to "`package test_test`" works. Just an example to improve this... we should rename the entire package name from other nontest files to "fixtures" and use in test files something like this "`package fixtures_test`".

Let me know if it's clear.

LandonTClipp: Ah! Okay, that's interesting. I think I'll want to change how the fixtures are structured, but I think your solution is good for now. Thanks.

LandonTClipp (approved):
This looks great. I'll approve for now and give it one last look later today and merge once the tests pass. Thanks for the contribution ❤️

mateusmarquezini:
Awesome! Thanks for your patience and support @LandonTClipp!

PR #750, opened 2024-01-24
Add ability to replace type constraints

### LINKED ISSUE #747: Option to replace a generic type parameter with a specific type
Description
-------------

When using an interface that describes a method that returns a clone of the struct, the generated mocks cannot be used.

Take this interface for example:

```go
type Repository[T any] interface {
  Clone() T
}
```

This can be implemented and used normally with this struct (note, no generics used here):

```go
type repository struct {}

func (r *repository) Clone() *repository {
  return &repository{}
}
```

This struct can successfully be injected into a function that expects the Repository[T] interface:

```go
func Foo[R Repository[R]](repo R) {
  // ...
}

func main() {
  repository := &repository{}
  Foo(repository)
}
```

However, the generated mocks generate a struct that follows the interface (as expected):

```go
type MockRepository[T interface{}] struct {
  mock.Mock
}

func (_m *MockRepository[T]) Clone() T {
  // ...
}
```

The problem is that these generated structs now cannot be used in the Foo function because their clone methods cannot infer T.

For an example with a little bit more added, please see this repository: https://github.com/DustinJSilk/mockery-cloneable/tree/main

Mockery Version
--------

v2.39.2

Golang Version
--------------

1.21.6

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [X] brew
- [ ] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. Clone the repository
2. See the errors in your IDE

### Expected Behavior

There is no way to have an interface that clones itself without using generics.
It would help if there was a way to tell Mockery to replace the generics with a specific type, in this case, itself.

### Actual Behavior

Is correct, but also doesnt work. It generates the mocks usings the interface definition but it needs more flexibility to work correctly.

DustinJSilk:
Description
-------------

Adds a config option that can replace generic types with a concrete type.

- Fixes #747

## Type of change

- [X] New feature (non-breaking change which adds functionality)

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [ ] 1.17
- [ ] 1.18
- [ ] 1.19
- [ ] 1.20
- [X] 1.21

How Has This Been Tested?
---------------------------

Checklist
-----------

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation


LandonTClipp:
Excellent! I'll give it a proper review in a few days. Thanks for the contribution.

pkg/generator.go
```diff
@@ -461,6 +473,10 @@ type namer interface {
 }
 
 func (g *Generator) renderType(ctx context.Context, typ types.Type) string {
+	if replacement, ok := g.config.ReplaceGeneric[typ.String()]; ok {
+		return replacement
+	}
+
```
LandonTClipp, line 479: To be semantically correct, what this is actually doing is to replace any type (not just generics). I wonder if we should instead change the name of the parameter to be something like `replace-param-type`?

I need to think about this carefully because we already have `replace-types` which _should_ be able to handle this, but that feature I don't think works with generics. I'm open to ideas.

DustinJSilk: You're right, it should use replace-types. I'll update the PR this week!

DustinJSilk:
Hi @LandonTClipp 

I've updated the API to the suggestion above which I'm using successfully for my usecase and hopefully it provides more flexibility for others.

I'm happy to change the API if you have any ideas.

LandonTClipp:
Hi @DustinJSilk, apologies for the lack of responses the last few weeks. I have been on the hunt for a new job and I haven't been able to devote much time or energy to this project, but luckily my search is coming to a close (hopefully, knock on wood).

As far as your proposals, I think I'm on board with it. The only thing I'm unhappy about (which has nothing to do with your proposal or implementation) is that I don't enjoy these complex, custom parameter strings. I think in hindsight, I would have preferred `replace-types` to simply be a list of maps with individual parameters, instead of having to parse this string here. That ship has already sailed... it's something I'll probably deprecate once a better implementation is available.

docs/features.md
```diff
@@ -90,6 +90,24 @@ func (_m *Handler) HandleMessage(m pubsub.Message) error {
...
+```shell
+mockery --replace-type github.com/vektra/mockery/v2/baz/internal/foo.InternalBaz[-T]=baz:github.com/vektra/mockery/v2/baz.Baz
+```
+
+When replacing a generic constraint, you can replace the type with a pointer by adding a '*' before the output type name.
+
+```shell
+mockery --replace-type github.com/vektra/mockery/v2/baz/internal/foo.InternalBaz[-T]=baz:github.com/vektra/mockery/v2/baz.*Baz
+```
+
```
LandonTClipp, line 110: I think it would also be helpful to show code examples of what this would translate to.

pkg/generator.go
```diff
@@ -1105,6 +1160,19 @@ func parseReplaceType(t string) *replaceType {
 		ret.alias = r[0]
 		t = r[1]
 	}
+
+	// Match type parameter substitution
+	match := regexp.MustCompile(`\[(.*?)\]$`).FindStringSubmatch(t)
+	if len(match) >= 2 {
+		if match[1][:1] == "-" {
```
LandonTClipp, line 1167: I was initially confused why you did that instead of:

```go
match[1][0]
```

but realized the range operator will return a string instead of a single byte. To make this cleaner, I'd instead do this:

```go
if strings.HasPrefix(match[1], "-")
```

It's a bit more obvious what's happening.

pkg/generator_test.go
```diff
@@ -760,6 +760,32 @@ func (s *GeneratorSuite) TestReplaceTypePackageMultiple() {
...
+	})
+
+	s.checkGenerationRegexWithConfig("generic.go", "ReplaceGenericSelf", cfg, []regexpExpected{
+		// type ReplaceGenericSelf struct
+		{true, regexp.MustCompile(`type ReplaceGenericSelf struct`)},
+		// func (_m *ReplaceGenericSelf) A() *ReplaceGenericSelf
+		{true, regexp.MustCompile(`func \(_m \*ReplaceGenericSelf\) A\(\) \*ReplaceGenericSelf`)},
+	})
+}
+
```
LandonTClipp, line 788: I do not like these kinds of tests and am asking people not to do them. Any changes to the formatting can/will cause these kinds of tests to fail. Instead I prefer if we just create a new test file, probably in `pkg/fixtures/test` that checks these various scenarios through a functional test, rather than string comparison. You can just rely on the compiler to fail if the param types are mismatched, or maybe use type assertion through an empty interface in the test (actually, type assertion might be better in this case so that we don't cause all the tests to fail if the mock implementation is wrong somehow).

LandonTClipp (changes requested):
Looks great, I think I'm okay with the new format here. I have a few comments, if we can get those addressed then I will merge it!

pkg/generator.go
DustinJSilk: Let me know if it still makes sense for you, i was able to put it in one line with `strings.CutPrefix`

pkg/generator_test.go
DustinJSilk: Good idea! I've added some better test cases and removed these. I didn't quite figure out exactly how we could test just with type assertions, at the moment the compiler would complain if the types no longer match

LandonTClipp (approved):
Excellent, this looks good to me. Thanks again!

LandonTClipp:
Good work @DustinJSilk, feature has been released.

DustinJSilk:
Amazing thanks @LandonTClipp

LandonTClipp:
A note about v3: I'm having to drop this feature because the implementation in this PR uses the `Generator` struct which is being cut entirely in favor of the TemplateGenerator. If this functionality is needed in v3, it should be easy to port but it will have to be redone in the new templating scheme.

PR #773, opened 2024-04-17
Added InterfaceFile

istrau2:
Description
-------------

Added InterfaceFile to available fields in .mockery.yaml template

I would like to name the mock file similarly to the interface file. Currently there is no way to know the interface file name in the template.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/pkg/outputter.go
+++ b/pkg/outputter.go
@@ -210,6 +210,7 @@ func parseConfigTemplates(ctx context.Context, c *config.Config, iface *Interfac
 	data := struct {
 		InterfaceDir            string
 		InterfaceDirRelative    string
+		InterfaceFile           string
 		InterfaceName           string
 		InterfaceNameCamel      string
 		InterfaceNameLowerCamel string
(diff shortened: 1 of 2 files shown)


LandonTClipp:
Hey, thanks for the PR! This parameter would be potentially misleading because you might have multiple interfaces within a single file. People have asked this before and I mention it only works in the case where you have one interface per file.

istrau2:
@LandonTClipp Thanks for the quick response.

I understand the concern, however, as far as this package is concerned I don't see how this is misleading. It is in fact the name of the file that the interface is in. If the consumer has decided to name their mock files based on the original name, they should ensure that there is only one interface in each file. Otherwise, as I understand, it would be overwritten, that seems fairly clear (and frankly, would be immediately observed since the tests would break).

For us, we have been happily using mockery for a couple of years now. Recently we've been getting deprecation warnings and I've been trying to upgrade to the new configuration. We do in fact have one interface per file and use a file name like `interface_file_mock.go` alongside the original `interface_file.go` all throughout the codebase. I would like to upgrade in a backwards compatible manner (rather than having to rename/restructure the codebase).

Perhaps a warning is in order if two interfaces share the same file name (although I am not sure how to implement that)?

Thanks again

LandonTClipp:
I've thought about this a bit more and in order to help you migrate to the new config, I think we can go ahead with it. My main concern is that it causes confusion without making an explicit note in the docs about what the behavior is when multiple interfaces are in a single file. I will just make a note in the docs to make this clear.

LandonTClipp:
@istrau2 could you please add a test for this to ensure that the rendered template matches what we expect? Once you get that, I can approve.

istrau2:
@LandonTClipp Much appreciated, thanks a lot. I have added the test.

istrau2:
@LandonTClipp thoughts?

LandonTClipp:
Looks great!

PR #782, opened 2024-06-05
Skip RunAndReturn() on functions with no returns

### LINKED ISSUE #506: Mock return value provider function does not execute when the function doesn't return anything
(Software Versions)
```
% go version       
go version go1.18.2 darwin/amd64

% mockery --version
09 Sep 22 10:36 EDT INF Starting mockery dry-run=false version=v2.14.0
v2.14.0
```

(Description)
Mock return value provider function does not execute when the function doesn't return anything (returns void). The function will be marked as called (as defined by the `AssertNumberOfCalls` method), but its commands won't actually be executed.

(Sample Source Files)

Create this 'sample.go' file in the 'sample' directory.

```
package sample

/*
Generate mock structs of these two interfaces with the following two commands -

mockery --dir=sample --filename=StructWithVoidReturn.go --output=mocks --outpkg=mocks --name=StructWithVoidReturn
mockery --dir=sample --filename=StructWithErrorReturn.go --output=mocks --outpkg=mocks --name=StructWithErrorReturn

*/
type StructWithVoidReturn interface {
	SomeCallWithVoidReturn()
}
type StructWithErrorReturn interface {
	SomeCallWithErrorReturn() error
}
```

Note - As the comment in this file indicates, create the two mock files as indicated in the block comment within the above file.

(Sample Test Files)

Create this 'sample_test.go' file in the 'sample' directory.

```
package sample

import (
	"testing"

	"github.com/iStreamPlanet/mediaflow-base/v2/mocks"
)

func TestStructWithVoidReturn(t *testing.T) {
	structWithVoidReturn := mocks.StructWithVoidReturn{}
	structWithVoidReturn.On("SomeCallWithVoidReturn").Return(
		func() {
			t.Logf("// 'SomeCallWithVoidReturn' mock function has been called")
		},
	)
	structWithVoidReturn.SomeCallWithVoidReturn()
	structWithVoidReturn.AssertNumberOfCalls(t, "SomeCallWithVoidReturn", 1)
}

func TestStructWithErrorReturn(t *testing.T) {
	structWithErrorReturn := mocks.StructWithErrorReturn{}
	structWithErrorReturn.On("SomeCallWithErrorReturn").Return(
		func() error {
			t.Logf("// 'SomeCallWithErrorReturn' mock function has been called")
			return nil
		},
	)
	structWithErrorReturn.SomeCallWithErrorReturn()
	structWithErrorReturn.AssertNumberOfCalls(t, "SomeCallWithErrorReturn", 1)
}

```

(Sample Execution)
Now run command `go test ./... -tags=unit_test -v` .

(Supporting Evidence)
Notice that `// 'SomeCallWithErrorReturn' mock function has been called` is present in the output but `// 'SomeCallWithVoidReturn' mock function has been called` is not. Also notice that these two test functions successfully call `AssertNumberOfCalls` to confirm that the respective function that each is testing has been called exactly once.

(Supplemental Evidence)
Studying the mock struct files themselves reveals that they have a different implementation for their respective function.

//StructWithVoidReturn.go
```
// SomeCallWithVoidReturn provides a mock function with given fields:
func (_m *StructWithVoidReturn) SomeCallWithVoidReturn() {
	_m.Called()
}
```

//StructWithErrorReturn.go
```
// SomeCallWithErrorReturn provides a mock function with given fields:
func (_m *StructWithErrorReturn) SomeCallWithErrorReturn() error {
	ret := _m.Called()

	var r0 error
	if rf, ok := ret.Get(0).(func() error); ok {
		r0 = rf()
	} else {
		r0 = ret.Error(0)
	}

	return r0
}
```
Note how the `SomeCallWithVoidReturn` function does not contain the if-else block nor does it contain a `rf()` command like the `SomeCallWithErrorReturn` function.

skitt:
Description
-------------

On functions that don't return anything, RunAndReturn() doesn't call the function given. To avoid misleading users, this changes mockery to not generate RunAndReturn() on such functions; it doesn't seem particularly useful anyway because there's no return value to specify.

Return() is preserved even though it isn't needed either, to avoid confusion with mock.Call.

- Fixes #506

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This is technically a breaking change, but it breaks broken code, so on balance it might be better!

- [x] 1.22

How Has This Been Tested?
---------------------------

I’ve updated the test suite to account for the expected change.

Checklist
-----------

### DIFF
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Expecter.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Expecter.go
@@ -149,11 +149,6 @@ func (_c *Expecter_NoReturn_Call) Return() *Expecter_NoReturn_Call {
 	return _c
 }
 
-func (_c *Expecter_NoReturn_Call) RunAndReturn(run func(string)) *Expecter_NoReturn_Call {
-	_c.Call.Return(run)
-	return _c
-}
-
 // Variadic provides a mock function with given fields: ints
 func (_m *Expecter) Variadic(ints ...int) error {
 	_va := make([]interface{}, len(ints))
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/ExpecterAndRolledVariadic.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/ExpecterAndRolledVariadic.go
@@ -149,11 +149,6 @@ func (_c *ExpecterAndRolledVariadic_NoReturn_Call) Return() *ExpecterAndRolledVa
 	return _c
 }
 
-func (_c *ExpecterAndRolledVariadic_NoReturn_Call) RunAndReturn(run func(string)) *ExpecterAndRolledVariadic_NoReturn_Call {
-	_c.Call.Return(run)
-	return _c
-}
-
 // Variadic provides a mock function with given fields: ints
 func (_m *ExpecterAndRolledVariadic) Variadic(ints ...int) error {
 	var tmpRet mock.Arguments
--- a/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Fooer.go
+++ b/mocks/github.com/vektra/mockery/v2/pkg/fixtures/Fooer.go
@@ -45,11 +45,6 @@ func (_c *Fooer_Bar_Call) Return() *Fooer_Bar_Call {
 	return _c
 }
 
-func (_c *Fooer_Bar_Call) RunAndReturn(run func(func([]int))) *Fooer_Bar_Call {
-	_c.Call.Return(run)
-	return _c
-}
-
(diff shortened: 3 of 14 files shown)


LandonTClipp:
Thanks for the PR. I'm not sure I want to just remove this. I feel like it would be better to change the mock so that it actually calls the provided function. From a user's perspective, when I use this function I am expecting the function to be executed, but obviously I wouldn't expect any return value. The proposal here of just removing `RunAndReturn` would make me upset as I now have to modify all my tests that use this. Instead, I would have preferred mockery to do what I expect in this scenario, which is to run the function.

I know that it will become redundant with `.Run` but I think I'm okay with that. I'd prefer the mock methods to be consistent rather than strictly orthogonal.

skitt:
Right, that makes sense.

PR #835, opened 2024-10-28
fix(generator): support nil returns when using interface typed params

### LINKED ISSUE #513: Bug: Cannot return nil for generic type T, when T is an interface
Description
-------------

For the interfaces
```golang
type Builder[T any] interface {
	Build() (T, error)
}

type Thing interface {
	DoStuff()
}
```
I'd like to mock `Builder` behavior as
```golang
thingBuilder := new(mocks.Builder[Thing])
thingBuilder.On("Build").Return(nil, errors.New("some error"))
```

But I get an error like:
```
panic: interface conversion: interface is nil, not Thing
```

Shouldn't mockery know that `Thing` is an interface and so `Builder[Thing].Build()` should be able to return `(nil, error)`?

Mockery Version
--------
``` bash
$ mockery --version
19 Oct 22 17:49 MDT INF Starting mockery dry-run=false version=v2.14.0
v2.14.0
```

Golang Version
--------------
``` bash
$ go version
go version go1.19 linux/amd64
```

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [x] go install
- [ ] Other: [specify]

### Expected Behavior

Can mock generic return types as nil when they are in fact interfaces.

### Actual Behavior

Tests panic

jfragosoperez:
Description
-------------

- Types using generic typed interface params, should allow nil values on return. 
- Some test/example mocks haven't been regenerated recently, so 'task mock' removed and recreated mocks taking into account the yaml config and package movements, etc. 

- Fixes #513 

## Type of change

- [ ] 1.22
- [X] 1.23

How Has This Been Tested?
---------------------------

- Added test coverage 
- Regenerate test mocks to reflect the changes (NOTE: interface{} renamed to any by default for new golang versions) 
```bash
task mocks
task test
```

Checklist
-----------


LandonTClipp:
Thank you for the PR, this is on my radar so I will give it a review once I find some spare time... hopefully very soon 👍🏻

jfragosoperez:
@LandonTClipp cool, I'll wait for your news, in case any changes were required, I'll apply them ASAP. Thanks

pkg/generator_test.go
```diff
@@ -337,6 +337,22 @@ func (s *GeneratorSuite) TestGeneratorChecksInterfacesForNilable() {
...
+
+	cfg := GeneratorConfig{
+		StructName:     "RequesterIfaceWithInterfaceTypedParam",
+		WithExpecter:   true,
+		UnrollVariadic: false,
+	}
+
+	s.checkGenerationWithConfig("requester_iface.go", "RequesterIfaceWithInterfaceTypedParam", cfg, expected)
+}
+
```
LandonTClipp, line 355: Instead of this, I'd prefer to run an actual test on the generated mock itself. Like, run a test that checks that we can supply a nil value as the return and that we get the nil interface back. This test you have here is just doing a simple text comparison, which is not useful.

This might be a bit awkward if the generated mock is placed in the separate `mocks/` directory. Instead, generate it in the same path as the fixture itself to prevent cyclic imports.

LandonTClipp (changes requested):
Ah, a simple fix! Thank you. One comment about the test.

pkg/generator_test.go
jfragosoperez: Hello 👋  I've applied the changes in a separate commit so you're able to see what I've changed. Looking forward to hearing from you to know if it fulfills the purpose or should I apply any other small tweak

LandonTClipp (approved):
Wonderful, thank you!

jfragosoperez:
@LandonTClipp I rebase and squashed in a single commit, so you're able to merge (the workflow will need to run first though)

PR #843, opened 2024-11-19
Add `resolve-type-alias` parameter

### LINKED ISSUE #839: Circular dependency in go 1.23 when mock is used in the package where it's defined
Description
-------------

Hey, looks like #808 results in a circular dependency if mock is used in the package where it's defined:

```go
// ./p/iface.go
package p

type T = int

//go:generate mockery --quiet --name I
type I interface{ F(T) }

// this generates ./p/mocks/I.go
package mocks
import "p"

func (_m *I) F(_a0 p.T) {
...

// ./p/i_test.go
import "p/mocks" // fails because of circular dependency
```

Here is a sample repo that reproduces the issue: https://github.com/max-melentyev/mockery-issue

Mockery Version
--------

2.46.3

Go Version
--------------

1.23

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [x] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. `git clone https://github.com/max-melentyev/mockery-issue`
2. `make test`

### Expected Behavior

Mock uses original type (`int`) rather than its local alias (`p.T`) and this allows using mock in the same package (`p`).

### Actual Behavior

`p/mocks` uses type `p.T` and this doesn't allow to use this mock from the package `p`:
```
go test -v ./p
# mockery-issue/p
package mockery-issue/p
        imports mockery-issue/p/mocks
        imports mockery-issue/p: import cycle not allowed in test
FAIL    mockery-issue/p [setup failed]
```

LandonTClipp:
Description
-------------

- Fixes #839


max-melentyev:
Do I understand correctly that this PR makes mocks to be created in `_test` package by default?
AFAIK `_test` packages are similar to `/internal/` and their content is not visible from other packages, even from other tests.

LandonTClipp:
Not quite. The relevant part of this PR is here: https://github.com/vektra/mockery/pull/843/files#diff-9233145a2afa8ecc02f5d52807d154baf4d2e2a118f22721cb80277f9b2e6d82R541-R553 The `.mockery.yaml` changes were just some cleanup I was doing.

You can see the two tests that demonstrate the new behavior: https://github.com/vektra/mockery/pull/843/files#diff-f52d7d97345ebe96212f5c1dca1585a844ef7af865ec3273358bc9b6597c0c83

max-melentyev:
@LandonTClipp looks like it now expands type to a struct definition:
```diff
-// Code generated by mockery v2.44.1. DO NOT EDIT.
+// Code generated by mockery v2.49.0. DO NOT EDIT.
....
-func (_m *CassandraAgentClientPool) GetClientForInstance(_a0 *manualv1alpha1.Instance) (pb.CassandraAgentClient, error) {
+func (_m *CassandraAgentClientPool) GetClientForInstance(_a0 *struct {
+       v1.TypeMeta
+       v1.ObjectMeta
+       Spec   manualv1alpha1.InstanceSpec   `json:"spec"`
+       Status manualv1alpha1.InstanceStatus `json:"status,omitempty"`
+}) (pb.CassandraAgentClient, error) {
```
Is this expected?

LandonTClipp:
Can you show the definition of your type?

max-melentyev:
```go
import (
        // ...
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)
// ...
type Instance struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   InstanceSpec   `json:"spec"`
	Status InstanceStatus `json:"status,omitempty"`
}
```

LandonTClipp:
I'm not able to replicate this behavior when using the `packages` config. Keep in mind that your configuration style is deprecated and will be removed soon.

LandonTClipp:
I used your reproducer for the original issue (modifying the Makefile so I can use the latest mockery).

```go
package p

type T = int

type Instance struct {
	attribute int
}

//go:generate mockery --quiet --name I
type I interface{ F(T) }

//go:generate mockery --quiet --name I2
type I2 interface{ F(Instance) }

```

The generated mock file for I2:

```go
// Code generated by mockery v2.49.0. DO NOT EDIT.

package mocks

import (
	p "mockery-issue/p"

)

// I2 is an autogenerated mock type for the I2 type
type I2 struct {
	mock.Mock
}

// F provides a mock function with given fields: _a0
func (_m *I2) F(_a0 p.Instance) {
	_m.Called(_a0)
}

```

You can see it is not writing out the struct definition, so I'm unclear how this is happening for you.

max-melentyev:
Here I could reproduce it: https://github.com/max-melentyev/mockery-issue/commit/ee08ac90b71344a1309ea299b80fc6f75f655c0e

Aliased struct gets inlined.

LandonTClipp:
I'm able to reproduce it as well, working on a fix.

LandonTClipp:
The merged PR above should resolve the issue for you.

PR #857, opened 2024-12-03
Make the parser accept index list expressions

### LINKED ISSUE #856: Type declarations that are `*ast.IndexListExpr` are not accepted
Description
-------------

Let's consider the following code:
```go
package pkg

type Fooer[T any] interface {
    Foo(T)
}

type StringFooer Fooer[string]
```

`mockery` will generate both mocks, as expected. Note that `StringFooer`'s declaration is an `*ast.IndexExpr`, which is accepted when parsing as per [this line in `pkg/parse.go`](https://github.com/vektra/mockery/blob/3a3fb6e01f3ec5ba8903a925414aae058b317a1c/pkg/parse.go#L382).

However, when there is more than one type parameter:
```go
package pkg

type Fooer[T1 any, T2 any] interface {
    Foo(T1, T2)
}

type StringFooer Fooer[string, int]
```

`StringFooer`'s declaration is now an `*ast.IndexListExpr`, which is regarded as unacceptable, the interface is ignored and no mock is produced. I discovered the fact by running mockery with `--log-level debug`.

I tried the naive approach of adding `*ast.IndexListExpr` to the aforementioned case and produced a custom `mockery` binary that I then ran on the previous example and a working mock was generated.

I ran mockery e2e and unit tests after the change and all of them passed. I also used the mock generated previously in a (admittedly simple) test with successful results.

Is there a reason `*ast.IndexListExpr` is not accepted or was it just an overlook? I'm not super familiar with Go's `ast` package, so I may be missing something. I'd be more than happy to open a PR with the proposed change if it was just a mistake.

Mockery Version
--------

v2.49.1

Go Version
--------------

1.23.3

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [x] brew
- [ ] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. Create a source file such as:
```go
package pkg

type Fooer[T1 any, T2 any] interface {
    Foo(T1, T2)
}

type StringFooer Fooer[string, int]
```

2. Create a `.mockery.yml` file:
```yml
with-expecter: True
packages:
  path/to/pkg:
    interfaces:
      Fooer:
      StringFooer:
```

3. Run `mockery`

### Expected Behavior

I'd expect to get mocks for both `Fooer` and `StringFooer`.

### Actual Behavior

Only `Fooer` gets a mock created.

volmedo:
Description
-------------

Interfaces defined as instantiation of generic interfaces with more than one type parameter ([example](https://github.com/volmedo/mockery/blob/267dbdfe67d79c9a46cd76513734ec5e71ce80f4/pkg/fixtures/index_list_expression.go#L7)) are described as `*ast.IndexListExpr` by Go's AST library.

These interfaces were not accepted by the parser. This PR solves the issue by adding index list expressions as an accepted expression type.

- Fixes #856 

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

Added a new fixture illustrating the case and ran custom builds against it before and after the change, confirming the mock was produced as expected after the fix.

I also added a generator test for this specific case.

Checklist
-----------

- [ ] (N/A) I have made corresponding changes to the documentation


LandonTClipp (changes requested):
Thanks, this looks mostly good. Can you just change the tests to follow the pattern here: https://github.com/vektra/mockery/tree/master/pkg/fixtures/type_alias

Basically:

1. Create subpackage in fixtures directory
2. Create interface
3. [Change `.mockery.yaml`](https://github.com/vektra/mockery/blob/master/.mockery.yaml#L97) to place mocks in that directory.
4. Add tests in that directory.

This is the model I'm moving towards. If you can do that I'll merge. Thanks!

volmedo:
I moved the fixtures and mocks to a subpackage and added some tests there as per the instructions. Let me know if I missed something.

On a side note, having actual tests in a folder called "fixtures" feels a bit weird. I'd create a new "tests" folder and move subpackages with test code in the "fixtures" folder there. Just a nitpick, feel free to disregard 🙂.

LandonTClipp (approved):
Looks great, thanks! Yeah the naming is hard, there is a lot of historical cruft in this repo that I haven't been able to clean up. Maybe soon.

Issue #845, opened 2024-11-19
`outpkg` not respected when `inpackage: True` is set


LandonTClipp:
This happens in generator.go:

```go
// GeneratePrologue generates the prologue of the mock.
func (g *Generator) GeneratePrologue(ctx context.Context, pkg string) {
	g.populateImports(ctx)
	if g.config.InPackage {
		g.printf("package %s\n\n", g.iface.Pkg.Name())
	} else {
		g.printf("package %v\n\n", pkg)
	}

	g.generateImports(ctx)
	g.printf("\n")
}
```

The `outpkg` parameter is completely ignored in this case, which prevents us from giving the mocks a package name like `original_pkg_test`. I will create a parameter that allows us to override this behavior and always respect `outpkg`. Sadly, we have to retain backwards compatability.

PR #870, opened 2024-12-27
Fix: warning everytime when `issue-845-fix: True`.

### LINKED ISSUE #862: Only log issue-845-fix if outpkg and inpackage are set
PLEASE READ
-------------

DO NOT submit tickets without _first_ using the latest version of Go, clearing your local golang package cache, and re-building mockery using the _latest_ Go version and the _latest_ version of mockery. Please provide evidence this has been done in your issue. Failure to provide this evidence will likely result in your issue being closed.

Description
-------------

Mockery outputs a `WRN` about #845 even when not using `inpackage: True`. As far as I can tell, without that, the issue is not related so it is confusing to see a warning. While we could add `issue-845-fix: true` to the config, it makes it confusing since others that read the file won't understand what it means. I think it makes sense to only warn when both outpkg and inpackage are set.

Mockery Version
--------

2.5.0

Go Version
--------------

1.23.4

```
NOTE: Please upgrade to the latest golang version before submitting tickets!
```

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [ ] brew
- [X] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

1. [First Step]
2. [Second Step]
3. [etc]

### Expected Behavior

[what you expect to happen]

### Actual Behavior

[what actually happened]

Alhanaqtah:
Description
-------------

Now the issue #845 warning is displayed only if both the `inpackage` and `outpkg` parameters are set.

- Fixes #862

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

If both `inpackage` and `outpkg` parameters are set, a warning is displayed, otherwise there will be no warning. Tests and a linter were also launched.

Checklist
-----------


pkg/generator.go
```diff
@@ -423,14 +423,18 @@ func (g *Generator) generateImports(ctx context.Context) {
 func (g *Generator) GeneratePrologue(ctx context.Context, pkg string) {
 	g.populateImports(ctx)
 
-	if !g.config.Issue845Fix {
+	// If both `inpackage` and `outpkg` are set warn
+	if g.config.InPackage && g.config.Outpkg != "" {
```
LandonTClipp, line 427: If I'm not mistaken, this deprecation warning will fire even if `Issue845Fix` is set to `true`.

LandonTClipp (changes requested):
Thanks for the PR! I think there is an issue with the logic.

pkg/generator.go
Alhanaqtah: I'm sorry, you're completely right.

Alhanaqtah:
@evanphx Could you approve this workflow, please)

PR #874, opened 2024-12-30
Fix empty build tags adding a tags parameter

RangelReale:
Description
-------------

`strings.Split` always returns at least 1 item in the result array, even for the empty string. This makes the `packages.Load` call to always have an empty `-tags` parameter.

- Fixes # (issue)

## Type of change

- [ ] 1.22
- [X] 1.23

How Has This Been Tested?
---------------------------

The debug log output showed the empty `-tags` parameter.

Checklist
-----------


LandonTClipp:
Good catch, thanks 👍🏻

PR #908, opened 2025-01-31
docs: fix go mod download command

ravsii:
Description
-------------

There's no such command as `go download`. It's `go mod download`


LandonTClipp:
Oops! Thank you.

PR #911, opened 2025-02-01
feat(v3): Warning for nonexistent mocks

### LINKED ISSUE #905: Getting an error when mock does not exist?
Let's say I'm using a config like this: 
```yaml
packages:
  pkg1/pkg2:
    interfaces:
      Interface1:
      Interface2:
```

Later on, we decided that we no longer need `Interface2`, but forgot to remove it from the config but mockery doesn't give us an error as it used to be. So there's no way to check whether that mock still being generated or ignored because the original interface wasn't found. I'm using mockery `v.2.51.1`

Haven't found anything related in the [docs](https://vektra.github.io/mockery/latest/configuration/)

In our project we do have a lot of mocks, but we pick them by hand (`all: false` && `recursive: false`). Can we add some kind of an option like `ignore-not-found` so it's possible to set it to false and get an error?

Here's how it used to be with the CLI generation, we explicitly get an error on such case:
```
go install github.com/vektra/mockery/v2@v2.46.2
mockery --dir ./pkg1/pkg2--inpackage --name=Interface2
27 Jan 25 10:23 MSK INF couldn't read any config file version=v2.46.2
27 Jan 25 10:23 MSK INF Starting mockery dry-run=false version=v2.46.2
27 Jan 25 10:23 MSK INF Using config:  dry-run=false version=v2.46.2
27 Jan 25 10:23 MSK WRN DEPRECATION: use of the packages config will be the only way to generate mocks in v3. Please migrate your config to use the packages feature. dry-run=false migration=https://vektra.github.
io/mockery/v2.46/migrating_to_packages/ url=https://vektra.github.io/mockery/v2.46/features/#packages-configuration version=v2.46.2
27 Jan 25 10:23 MSK INF Walking dry-run=false version=v2.46.2
27 Jan 25 10:23 MSK ERR Unable to find 'Interface2' in any go files under this path dry-run=false version=v2.46.2
unable to find interface
```

ravsii:
Description
-------------

Copy of #907 but for v3. Also, no `log.Ctx()` this time :D

- Fixes #905 for v3

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

```
2025-02-01T19:46:07.124329000+03:00 WRN no such interface interface=InterfaceDoesntExist package-path=github.com/vektra/mockery/v3/internal version=v0.0.0-dev
```

Checklist
-----------

---

_On a side note, don’t you think that RootApp.Run is a little too big? Having 2 (or 3 now) huge loops with all the core logic feels like a bit too much. Just my two cents as an outside contributor._

_I've also put my code in there, but looks like it could become a mess with more features_

### DIFF
--- a/internal/cmd/mockery.go
+++ b/internal/cmd/mockery.go
@@ -7,14 +7,15 @@ import (
 	"os"
 	"strings"
 
-	"github.com/chigopher/pathlib"
-	"github.com/rs/zerolog"
-	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"github.com/vektra/mockery/v3/config"
 	pkg "github.com/vektra/mockery/v3/internal"
 	"github.com/vektra/mockery/v3/internal/logging"
 	"github.com/vektra/mockery/v3/internal/stackerr"
+
+	"github.com/chigopher/pathlib"
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"golang.org/x/tools/go/packages"
 )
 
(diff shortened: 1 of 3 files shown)


internal/cmd/mockery.go
```diff
@@ -322,5 +356,16 @@ func (r *RootApp) Run() error {
 		}
 	}
 
+	// The loop above could exit early, so sometimes warnings won't be shown
+	// until other errors are fixed
+	for packageName := range missingMap {
```
LandonTClipp, line 361:
```suggestion
	for packagePath := range missingMap {
```

Technically, a "package name" is different from a "package path".

internal/cmd/mockery.go
```diff
@@ -322,5 +356,16 @@ func (r *RootApp) Run() error {
...
+		for iface := range missingMap[packageName] {
+			log.Warn().
+				Str(logging.LogKeyInterface, iface).
+				Str(logging.LogKeyPackagePath, packageName).
+				Msg("no such interface")
```
LandonTClipp, line 366:
```suggestion
				Msg("interface not found in source")
```

Just to be a bit more explicit.

LandonTClipp (changes requested):
Just a couple of tweaks and this should be good to go. Thanks for the PR!

Regarding your comment about the length of `.Run()`, I don't necessarily believe that functions need to be split up because of line counts. My philosophy is that you should break it up for any of the following reasons:

1. To make unit testing easier.
2. To factor out common bits of logic used in more than one place (or in other words, abstracting complex bits of logic out).
3. To make it easier to understand what is happening.

There are little to no unit tests for this method because I'm relying much more heavily on e2e integration tests, so point 1 is moot. There is only one consumer of this logic, so point 2 is moot. And IMO the method is fairly easy to understand so point 3 is moot. I would prefer to have all the core business logic in one large-ish function if it makes it easier to follow the flow of the logic. Abstractions can sometimes run the risk of making it harder to know what's actually happening.

I understand this is just a preference, but from a practical perspective, I don't see a reason to break it up. But, I appreciate the perspective!

internal/cmd/mockery.go
ravsii: makes sense, fixed

internal/cmd/mockery.go
ravsii: sure

LandonTClipp (approved):
Looks great! Thanks again.

PR #918, opened 2025-02-13
tests(v3): add tests for template funcs

### LINKED ISSUE #916: v3: Extend tests for template funcs to include shared funcs
We have a list of template funcs here: https://github.com/vektra/mockery/blob/v3/shared/template_funcs.go

These are shared between the template rendering itself and the configuration option templates. These were tested in v2 but they have no corresponding unit tests in v3. https://github.com/vektra/mockery/blob/v2.52.1/pkg/outputter_test.go#L120-L140

The reason I noticed this is because one of the packages used in the shared template funcs (xstrings) introduced breaking API changes. We need to port the unit tests over to v3 to protect against this.

ravsii:
Description
-------------

Added tests for template functions, did some refactor of existing ones as well.

- Fixes #916

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

Added tests

Checklist
-----------

### DIFF
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -2,52 +2,343 @@ package template
 
 import (
 	"go/types"
+	"os"
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+func TestTemplateStringFuncs(t *testing.T) {
+	// For env tests
+	os.Setenv("MOCKERY_TEST_ENV", "TEST")
+
+	tests := []struct {
+		name       string
+		inTemplate string
+		inData     map[string]any
+		want       string
+	}{
+		{
+			name:       "contains",
+			inTemplate: "{{contains .TemplateData.sub .TemplateData.str}}",
+			inData:     map[string]any{"str": "golang", "sub": "go"},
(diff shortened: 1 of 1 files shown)


template/template_test.go
```diff
@@ -2,52 +2,347 @@ package template
...
+			want: "true",
+		},
+		{
+			name:       "typeConstraint",
+			inTemplate: "{{index .Mocks 0 | typeConstraintTest}}",
+			dataInit: func() Data {
+				return Data{Mocks: []MockData{{
+					TypeParams: []TypeParamData{{
+						ParamData: ParamData{
+							Var: &Var{Name: "t", typ: &types.Slice{}},
```
ravsii, line 308: couldn't find a better type that implements that interface, and I can't change the underlying type of a slice element, so it's `[]<nil>` in the tests

ravsii:
Also, I think it's time to add 1.24 to the pipeline

LandonTClipp:
Awesome this is great! Thanks. I will take a closer look later.

LandonTClipp (approved):
Really great, no comments at all. Love how clean it all is :D 

Thanks again.

PR #926, opened 2025-02-23
Fix profiling

mat007:
Description
-------------

This PR fixes the `--profile` flag, mainly by moving it above the
```
	if len(configuredPackages) == 0 {
…
	} else {
…
	}
```
block.

Also the `--cpuprofile` flag was not linked to any config field (anymore?).
I think there was possibly a renaming at some point, and it broke it?

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

Checklist
-----------

### DIFF
--- a/cmd/mockery.go
+++ b/cmd/mockery.go
@@ -66,6 +66,7 @@ func NewRootCmd() *cobra.Command {
 	pFlags.Bool("testonly", false, "generate a mock in a _test.go file")
 	pFlags.String("case", "", "name the mocked file using casing convention [camel, snake, underscore]")
 	pFlags.String("note", "", "comment to insert into prologue of each generated file")
+	pFlags.String("cpuprofile", "", "write cpu profile to file")
 	pFlags.String("profile", "", "write cpu profile to file")
 	pFlags.Bool("version", false, "prints the installed version of mockery")
 	pFlags.Bool("quiet", false, `suppresses logger output (equivalent to --log-level="")`)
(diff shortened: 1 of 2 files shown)


cmd/mockery.go
```diff
@@ -66,7 +66,7 @@ func NewRootCmd() *cobra.Command {
 	pFlags.Bool("testonly", false, "generate a mock in a _test.go file")
 	pFlags.String("case", "", "name the mocked file using casing convention [camel, snake, underscore]")
 	pFlags.String("note", "", "comment to insert into prologue of each generated file")
-	pFlags.String("cpuprofile", "", "write cpu profile to file")
+	pFlags.String("profile", "", "write cpu profile to file")
```
LandonTClipp, line 69: We can't change the param name as that's a breaking change, even if it's not linked to anything. I looked back at v1 and it seems this was originally tied to the `--cpuprofile` flag: https://github.com/vektra/mockery/blob/v1.1.2/cmd/mockery/mockery.go#L152

I must have broken this during the v2 revamp and no one noticed until now :D

Probably what we should do is leave the `--cpuprofile` param, but check for `if r.Config.Profile || r.Config.Cpuprofile` because `profile` could still have been set in config or the environment.

LandonTClipp (changes requested):
Thanks for catching this, I'm surprised no one complained for (checks notes) 5 years.

cmd/mockery.go
mat007: All right, I added a commit to add back `--cpuprofile`.

LandonTClipp (approved):

cmd/mockery.go
LandonTClipp: Best to leave `--profile` off of the parameter list entirely, actually. It can still be set via config.

LandonTClipp (changes requested):

cmd/mockery.go
mat007: Done.

LandonTClipp (approved):

LandonTClipp:
Thank you!

PR #928, opened 2025-02-24
fix: `include-auto-generated` option

### LINKED ISSUE #920: `include-auto-generated: false` seems ignored
I have an `api` package, that contains both OpenAPI definitions (generated using [ogen](https://github.com/ogen-go/ogen)) and actual implementations for those definitions (in separate files).

Ny mockery configuration looks like this

```yaml
packages:
  github.com/my-org/my-module/api:
    config:
      include-regex: '^[A-Z].+'
      include-auto-generated: false
      recursive: false
      outpkg: apimocks
      dir: api/mocks
```

# Actual behavior

When I run `go run github.com/vektra/mockery/v2@v2.52.2`, it picks up the ogen definitions, even though they belong to generated files (`// Code generated by ogen, DO NOT EDIT.` at the top).

I don't want those definitions picked up because:

- I don't use them in the interfaces I want to test (it's the opposite: I implement them)
- ogen generation includes exported interfaces with private methods, that take private (unexported) types as arguments. Mockery files are broken because of this (since they can't import private types).

# Expected behavior

If I set `include-auto-generated: false`, generated files should not be picked up.

ravsii:
Description
-------------

This fixes `include-auto-generated` option that was ignored for some time.

To be honest the fix is very far from optimal, because Ideally, we'd like to ignore the generated file once, but that's not possible at the moment, because "per-file" access is done inside `parse.go`, but we can't import `config` there for package config, so the only sensible place for that was `mockery.go`. Because of that we check all the interfaces (even auto-generated ones) for the magic DO NOT EDIT inside their files.\
I didn't want to fight dependencies too much, so I've done it in a very straightforward & nonoptimal way. Otherwise we'd probably have to let parser have access to config somehow, without making it a circular import.

Publishing it as a Draft first, because I still need to do E2E scripts

- Fixes #920

## Type of change

- [x] 1.24

How Has This Been Tested?
---------------------------

Checklist
-----------


ravsii:
@LandonTClipp Hi, could you check please if this is something you've expected or there's a better way to do this?

LandonTClipp:
This is looking great so far. Regarding the issue of the parser not having access to the config object, we could probably get around that by having the parser accept an interface like 

```go
type shouldGenerate interface {
     ShouldGenerateInterface(ctx context.Context, packageName, interfaceName string) (bool, error)
}
```

in its constructor, and then pass `config.Config` in directly. That would (I hope) get around the circular dependency issue.

There might also be a caching opportunity here where you can populate a `map[Filename] bool` with files that are known to be auto-generated (or files that are known to not be auto-generated). Future interfaces can check if its file is in that map and use that result. Otherwise, parse all the comments.

What you have isn't actually all that terrible, it's definitely in the right direction. If you can make those small tweaks with the dependency injection, I think it'll be perfect. Also no rush, folks have gone 5 years without submitting a complaint about this :D

ravsii:
> There might also be a caching opportunity here where you can populate a `map[Filename] bool` with files that are known to be auto-generated (or files that are known to not be auto-generated). Future interfaces can check if its file is in that map and use that result. Otherwise, parse all the comments.

Yeah that's pretty much what I wanted to do, but as I said only parser have "per-file" access, later on it's only "per-interface", so not very optimal. Didn't want to refactor parser that much

I'll see what I can do, but interface seems nice

ravsii:
I don't know how high I was doing the first set of changes lol.

So after doing some more iterations, I realized the best solution would be to store "IsAutoGenerated" as a field of both interfaces and files, representing if that file/interface was, in face, auto generated.

So later on me and someone else could use it to implement any kind of logic they want, instead of trying to make parser be the "do it all" kind of object, which obviously it shouldn't be.

@LandonTClipp could you check it now, please?

pkg/config/config.go
```diff
@@ -292,12 +292,21 @@ func (c *Config) ExcludePath(path string) bool {
...
+	packageName, interfaceName string,
+	isAutoGenerated bool,
+) (bool, error) {
 	pkgConfig, err := c.GetPackageConfig(ctx, packageName)
 	if err != nil {
 		return false, fmt.Errorf("getting package config: %w", err)
 	}
 
+	shouldSkipGenerated := !pkgConfig.IncludeAutoGenerated
+	if shouldSkipGenerated && isAutoGenerated {
```
ravsii, line 306: compiler will optimize this into a single line, but for developers "shouldDoX" is much more readable instead of "give me the opposite of include" imo

LandonTClipp (approved):
I like the changes you made, this is a very clean implementation so well done! Approving 🎉

PR #940, opened 2025-03-06
fix: include-auto-generated option to work with multi-line comment

### LINKED ISSUE #939: include-auto-generated doesn't work for multi-line generation comments
I noticed that the include-auto-generated recently fixed in https://github.com/vektra/mockery/pull/928 doesn't respect multi-line auto-generation comments from the likes of sqlc:

```
// Code generated by sqlc. DO NOT EDIT.
//
//	sqlc v1.25.0
```

I found the culprit in the code and will provide a fix.

hpurmann:
Description
-------------

This further improves the fix made in
https://github.com/vektra/mockery/pull/928. That version of the fix didn't work for generated files with multi-line comment (e.g. sqlc).

Additionally it runs mockery in the e2e test.

- Fixes #939

## Type of change

- [ ] 1.22
- [ ] 1.23
- [x] 1.24

How Has This Been Tested?
---------------------------

Added unit test and improved e2e test.

Checklist
-----------


pkg/parse.go
```diff
@@ -18,7 +18,7 @@ import (
 	"golang.org/x/tools/go/packages"
 )
 
-var autoGeneratedRegex = regexp.MustCompile(`^\s*Code generated .* DO NOT EDIT\.(\n)?$`)
+var autoGeneratedRegex = regexp.MustCompile(`Code generated .* DO NOT EDIT\.`)
```
LandonTClipp, line 21: The way I would prefer to do this is instead loop over every line in the comment group.

https://github.com/vektra/mockery/blob/103f702b7d55c05bd13c9ee4dd25a60a9d31cac3/pkg/parse.go#L442

The [`ast.CommentGroup`](https://pkg.go.dev/go/ast#CommentGroup) can be iterated over, so it's probably better to leave this regex as-is and just run it against each specific `ast.Comment`.

LandonTClipp (changes requested):
Thanks for the PR, one minor request.

pkg/parse.go
hpurmann: Ok, I changed it. I now had to include the `//` in the regex, as the `Text()` function to strip it is only available on the `ast.CommentGroup`, not on individual `Comment` and I didn't want to re-implement this logic. I believe it should be fine as the generated comments always follow that format, e.g. are never using `/* ... */`

LandonTClipp (approved):
Looks good, thank you!

PR #941, opened 2025-03-10
Revert to Go 1.23 in go.mod

### LINKED ISSUE #937: What's the point of forcing go 1.24?
Hi, I wanted to update mockery to patch version 2.52.4 to fix some small issues, but it forces now golang 1.24. Isn't it a bit steep requirement for a patch version? Especially that nothing in the code uses that version?

p.s. The toolchain directive is probably unnecessary as well - `go 1.24.0` should be enough.

LandonTClipp:
It was found in #937 that the upgrade to Go 1.24 in go.mod was both unnecessary and also a breaking change. I made a dumb mistake in thinking the toolchain needed to be upgraded to support the generic type alias syntax introduced in Go 1.24, but all that was needed was upgrading `golang.org/x/tools` which is the utility that parses the syntax.

This fixes #937.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/vektra/mockery/v2
 
-go 1.24
-
-toolchain go1.24.0
+go 1.23
 
 require (
 	github.com/chigopher/pathlib v0.19.1
--- a/go.work
+++ b/go.work
@@ -1,6 +1,4 @@
-go 1.24.0
-
-toolchain go1.24.0
+go 1.23.0
 
 use (
 	.
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/vektra/mockery/tools
 
-go 1.24.0
+go 1.23
 
 require (
 	github.com/go-errors/errors v1.5.1


fredrikaverpil:
Hey, just wanted to mention that I saw this error when upgrading a project to Go 1.24 and pre-built Mockery `v2.53.1`:

```
10 Mar 25 15:01 CET ERR encountered error when loading package error="-: This application uses version go1.23 of the source-processing packages but runs version go1.24 of 'go list'. It may fail to process source files that rely on newer language features. If so, rebuild the application using a newer version of Go." dry-run=false version=v2.53.1
```

So it might still be the case that mockery is sensitive whether it is running in Go project of a different version than the one it was built with?

Anyway, by not using a pre-built `mockery` binary, and just do a `go install` instead, it works fine - without issues.

LandonTClipp:
Hmm, this may just be something we can fix in CI. As long as it builds with 1.24, this shouldn't happen. I'll look into it.

LandonTClipp:
@fredrikaverpil can you check whether [v2.53.2](https://github.com/vektra/mockery/releases/tag/v2.53.2) fixes your issue?

fredrikaverpil:
@LandonTClipp yes, the v2.53.2 pre-built binary seems to work now! 🎉

Issue #936, opened 2025-03-03
Using mockery V3 to generate https://github.com/hexdigest/gowrap style interface wrappers


breml:
I tried to use mockery V3 in the new matryer/moq mode to generate interface wrappers in the style of https://github.com/hexdigest/gowrap and I made it work, but I found some rough edges, that I would like to mention here:

* `gowrap` offers some helpers for usage in templates, that come in handy when generating interface wrappers, namely:
  * `.HasParams` and `.HasResults` on [`MethodData`](https://pkg.go.dev/github.com/vektra/mockery/v3/template#MethodData). With `mockery` I needed to write the condition `{{ if gt ($method.Returns | len) 0 }}` where in `gowrap` it is simply `{{ if $method.HasResults }}`
  * `.AcceptsContext` and `.ReturnsError` on [`MethodData`](https://pkg.go.dev/github.com/vektra/mockery/v3/template#MethodData). With `mockery` I needed create helper variables like this
    ```
    {{- $acceptsContext := false -}}
    {{- if and (gt ($method.Params | len) 0) (eq (index $method.Params 0).TypeString "context.Context") -}}
    {{- $acceptsContext = true -}}
    {{- end -}}
    ```
    and
    ```
    {{- $returnsError := false -}}
    {{- $lastType := "" -}}
    {{- range $method.Returns -}}
    {{- $lastType = .TypeString }}
    {{- end -}}
    {{- if eq $lastType "error" -}}
    {{- $returnsError = true -}}
    {{- end -}}
    ```
    in order to then be able to form conditions like [`{{- if $returnsError }}`](https://github.com/hexdigest/gowrap/blob/master/templates/log#L35), which is cumbersome.
    What made this additionally cumbersome is, that Go templates do no provide basic arithmetic in order to get the last return value with `($method.Returns | len) - 1` (this would be possible with e.g. the math functions provided by [sprig](https://masterminds.github.io/sprig/math.html)). Therefore I needed to range though the list only to get the last item.
* In contrast to mocks, wrappers (or middlewares) also perform a call to the thing, that they wrap. With gowrap, there is a single instruction called [`$method.Pass`](https://github.com/hexdigest/gowrap/blob/master/templates/log#L48), which accepts a prefix (basically the selector of the attribute of the wrapped instance). It returns the complete call to to the wrapped method including the arguments as well as the `return` statement, if the method does return anything. My workaround with `mockery` looks like this:
    ```
[...]

PR #960, opened 2025-03-29
Add a few convenience methods to MethodData

LandonTClipp:
This addresses some comments raised in #936. We add some methods to MethodData that make some common patterns in the templates easier to manage. Also added comments to matryer/testify templates that describe which template they are.

TODO: Add tests. Probably want to create a template file that exercises every part of `MethodData` (and possibly other parts of `template.Data`). Should do a 1-1 string comparison of the rendered template. This is necessary because we want to work off of real type data instead of mocking out any part of `template.Data`. Use `template: file//./path/to/test.templ` and run it as an e2e test.


LandonTClipp:
@breml I took your suggestions to heart and implemented a few of them. I followed your suggestions pretty much exactly with the exception of your `$method.Pass` comment. I looked into this and I felt that, stylistically, it leans a little too far into the realm of "doing too much" for the template. More specifically, I don't understand the benefit of the `prefix` argument when you can just add it in the template directly, and I also don't particularly like how it automatically manages the `return` statement for you. My reasoning on `return` is that templates are already so difficult to look at and abstract that keeping so-called "reality anchors" (which in this case is the `return` statement) helps understand what is actually happening in the template.

I admit this is purely a preference, and I'm open to being convinced otherwise. Let me know your thoughts.

breml:
One more idea about `$method.Pass`, maybe we can just add `$method.ReturnStatement`, such that we could write

```
{{ $method.ReturnStatement }} _d._base.{{ $method.Call }}
```

LandonTClipp:
That's great! That's awesome you were able to port the templates so easily. I'll have to keep those as a reference for when we add them to the project.

> What might be possible though is, that we change .Mocks (https://github.com/breml/mockery-wrap-test/blob/master/wrappers/fmt_print.gotmpl#L13) to something more generic (since we are still in beta). If you think this is too severe of a change that late in the process, then one other option would be to offer .Mocks with a second name. Basically we could just add a getter method (e.g. Interfaces) on https://github.com/vektra/mockery/blob/v3/template/template_data.go#L10, which returns .Mocks (basically serving as an alias). This would work nicely in templates, since there is no difference between methods and struct fields.

I'm not really decided yet on what to do here. It would be a pain to have to change the reference to "mock" everywhere, however I do prefer to do things the right way out of the gate because once we release to GA, there is no turning back. I'll have to think about this more.

> One more idea about $method.Pass, maybe we can just add $method.ReturnStatement, such that we could write
> {{ $method.ReturnStatement }} _d._base.{{ $method.Call }}

That's a great idea actually, I think it somewhat ameliorates my concern. I have to be clear though that this is not _really_ something I care deeply about. There's a lot more pressing concerns and this is not something to agonize over 😅 But, I think your suggestion is actually good enough for me so I'll add that in this PR.

PR #1022, opened 2025-04-19
Fix uppercase generic type parameters

### LINKED ISSUE #1021: Interface with lowercase generic type parameters gets converted to pascal case
Hello,

Using mockery `v3.2.3` and having the following interface:

```go
type Foo[baz comparable, goo comparable] interface {
	GetBaz(goo) baz
}
```

and using the default configuration the following code is being generated:

```go
// Code generated by mockery; DO NOT EDIT.
// github.com/vektra/mockery
// template: testify

package main

import (
)

// NewMockFoo creates a new instance of MockFoo. It also registers a testing interface on the mock and a cleanup function to assert the mocks expectations.
// The first argument is typically a *testing.T value.
func NewMockFoo[Baz comparable, Goo comparable](t interface {
	mock.TestingT
	Cleanup(func())
}) *MockFoo[Baz, Goo] {
	mock := &MockFoo[Baz, Goo]{}
	mock.Mock.Test(t)

	return mock
}

// MockFoo is an autogenerated mock type for the Foo type
type MockFoo[Baz comparable, Goo comparable] struct {
	mock.Mock
}

type MockFoo_Expecter[Baz comparable, Goo comparable] struct {
	mock *mock.Mock
}

func (_m *MockFoo[Baz, Goo]) EXPECT() *MockFoo_Expecter[Baz, Goo] {
	return &MockFoo_Expecter[Baz, Goo]{mock: &_m.Mock}
}

// GetBaz provides a mock function for the type MockFoo
func (_mock *MockFoo[Baz, Goo]) GetBaz(v goo) baz {
	ret := _mock.Called(v)

	if len(ret) == 0 {
		panic("no return value specified for GetBaz")
	}

	var r0 baz
	if returnFunc, ok := ret.Get(0).(func(goo) baz); ok {
		r0 = returnFunc(v)
	} else {
		if ret.Get(0) != nil {
			r0 = ret.Get(0).(baz)
		}
	}
	return r0
}

// MockFoo_GetBaz_Call is a *mock.Call that shadows Run/Return methods with type explicit version for method 'GetBaz'
type MockFoo_GetBaz_Call[Baz comparable, Goo comparable] struct {
	*mock.Call
}

// GetBaz is a helper method to define mock.On call
//   - v
func (_e *MockFoo_Expecter[Baz, Goo]) GetBaz(v interface{}) *MockFoo_GetBaz_Call[Baz, Goo] {
	return &MockFoo_GetBaz_Call[Baz, Goo]{Call: _e.mock.On("GetBaz", v)}
}

func (_c *MockFoo_GetBaz_Call[Baz, Goo]) Run(run func(v goo)) *MockFoo_GetBaz_Call[Baz, Goo] {
	_c.Call.Run(func(args mock.Arguments) {
		run(args[0].(goo))
	})
	return _c
}

func (_c *MockFoo_GetBaz_Call[Baz, Goo]) Return(v1 baz) *MockFoo_GetBaz_Call[Baz, Goo] {
	_c.Call.Return(v1)
	return _c
}

func (_c *MockFoo_GetBaz_Call[Baz, Goo]) RunAndReturn(run func(v goo) baz) *MockFoo_GetBaz_Call[Baz, Goo] {
	_c.Call.Return(run)
	return _c
}
```

This code doesn't compile, because all the type parameters are converted to pascal case:
- `baz` -> `Baz`
- `Goo` -> `Goo`

e.g. 

```go
func NewMockFoo[Baz comparable, Goo comparable](t interface {
	mock.TestingT
	Cleanup(func())
}) *MockFoo[Baz, Goo] {
	mock := &MockFoo[Baz, Goo]{}
	mock.Mock.Test(t)

	return mock
}
```

Should be:

```go
func NewMockFoo[baz comparable, goo comparable](t interface {
	mock.TestingT
	Cleanup(func())
}) *MockFoo[baz, goo] {
	mock := &MockFoo[baz, goo]{}
	mock.Mock.Test(t)

	return mock
}
```

and everywhere else in the generated code they are with lower-case. After looking at the code and the template, the following 2 functions are introducing this issue:

- https://github.com/vektra/mockery/blob/10fff726ef847b7091cd34854b1f19fe2effc7e4/template/interface.go#L54-L69
- https://github.com/vektra/mockery/blob/10fff726ef847b7091cd34854b1f19fe2effc7e4/template/interface.go#L71-L84

they both point to https://github.com/vektra/mockery/blob/10fff726ef847b7091cd34854b1f19fe2effc7e4/template_funcs/functions.go#L13-L23 and finally https://github.com/vektra/mockery/blob/10fff726ef847b7091cd34854b1f19fe2effc7e4/template_funcs/functions.go#L22

Is there a particular reason, why each of the type parameters need to be exported?

fen4o:
Description
-------------

The implementation doesn't modify the behavior of the existing exported function called `exported`.

- Fixes #1021

## Type of change

- [ ] 1.22
- [X] 1.23

How Has This Been Tested?
---------------------------

Checklist
-----------


template_funcs/functions.go
```diff
@@ -11,6 +11,14 @@ import (
...
 
 func Exported(s string) string {
+	s = ExportedTypeParameters(s)
+	if s == "" {
+		return ""
+	}
+	return strings.ToUpper(s[0:1]) + s[1:]
+}
+
+func ExportedTypeParameters(s string) string {
```
LandonTClipp: I would call this something other than `ExportedTypeParameters` because it's not actually exporting the type params. Actually, I would say we should _not_ export at all and instead just delete the `template_funcs.Exported(` lines from the `Interface` methods above. I'm considering this to be a bug so I'm comfortable making a backwards-incompatible change.

LandonTClipp (changes requested):
Thanks for the PR. I'd say just remove the `Exported(` lines entirely, I don't see a reason to keep this.

template_funcs/functions.go
fen4o: The main concern is that this function is exported to templates https://github.com/vektra/mockery/blob/10fff726ef847b7091cd34854b1f19fe2effc7e4/template_funcs/funcmap.go#L54 and if we remove it, custom templates which are using it will break.

fen4o: Also, looking at the code, I'm not sure what the idea behind https://github.com/vektra/mockery/blob/7714096a3897a387963f2173d14c7a47d3032833/template_funcs/funcmap.go#L18-L24 is. If you have those types in your code-base https://golangci-lint.run/ will probably complain in the first place. If you are generating mocks for some 3rd party library, you can already configure golang-ci to ignore generated mock files.

LandonTClipp: > The main concern is that this function is exported to templates

I'm not suggesting removing the functions, I'm suggesting removing these lines: https://github.com/vektra/mockery/pull/1022/files#diff-ecc1fcee461afe7551a8b4122c0e3b83718ebae30d68e8315d10fb8c0df7ccd7L63-R63 from the template. I don't think we need to be exporting the type parameter names at all.

fen4o: Ah I finally got you. Have another look

LandonTClipp (approved):
Looks great! Thanks.

The only possible issue we might run into in the future (which existed prior to your PR) is that the type parameter may not be going through naming collision resolution properly. I haven't looked closely to confirm, I'm just speculating. But, I think this will solve your original problem.

PR #1045, opened 2025-05-20
Fix missing dependency by running go mod tidy

miner46er:
Description
-------------

This fixes missing golang.org/x/exp dependency on install on some machine.

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

I've run `task test` and it passes.

Checklist
-----------


LandonTClipp:
Somehow this seems to affect the result of the template exercise test. We'll need to dig into why.

miner46er:
Looking at the last [known working pipeline](https://github.com/vektra/mockery/actions/runs/14769664382) I was seeing it fetches for `golang.org/x/exp` `v0.0.0-20240719175910-8a7402abbf56`, meanwhile the latest version of `golang.org/x/exp` is `v0.0.0-20250506013437-ce4c2cf36ca6` 

in `v0.0.0-20240719175910-8a7402abbf56` the definition of `constraints.Ordered` is ([link](https://github.com/golang/exp/blob/8a7402abbf56ed11a2540c1d8beb569bd29e22d1/constraints/constraints.go#L44)):
```go
// Ordered is a constraint that permits any ordered type: any type
// that supports the operators < <= >= >.
// If future releases of Go add new ordered types,
// this constraint will be modified to include them.
type Ordered interface {
	Integer | Float | ~string
}
```

while in `v0.0.0-20240719175910-8a7402abbf56` ([link](https://github.com/golang/exp/blob/master/constraints/constraints.go)):
```go
// Ordered is a constraint that permits any ordered type: any type
// that supports the operators < <= >= >.
// If future releases of Go add new ordered types,
// this constraint will be modified to include them.
//
// This type is redundant since Go 1.21 introduced [cmp.Ordered].
//
//go:fix inline
type Ordered = cmp.Ordered
```

and for further context, the definition of `cmp.Ordered` is ([link](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/cmp/cmp.go;drc=1a0b86375fad202048adb88cba4caec535a52a45;l=9)):
```go
// Ordered is a constraint that permits any ordered type: any type
// that supports the operators < <= >= >.
// If future releases of Go add new ordered types,
// this constraint will be modified to include them.
//
// Note that floating-point types may contain NaN ("not-a-number") values.
// An operator such as == or < will always report false when
// comparing a NaN value with any other value, NaN or not.
// See the [Compare] function for a consistent way to compare NaN values.
type Ordered interface {
	~int | ~int8 | ~int16 | ~int32 | ~int64 |
		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr |
		~float32 | ~float64 |
		~string
}
```

Since `constraints.Ordered` is used in template exercise fixture, that explains the difference in output.

Changing the `golang.org/x/exp` dependency to the former (`v0.0.0-20240719175910-8a7402abbf56`) fixes the issue.

Should I downgrade the dependency to that version or keep the latest version and update the test's expected output? @LandonTClipp

LandonTClipp:
Thanks for digging into that @miner46er ! I think we should downgrade the dependency to make the tests pass. It may be fine to upgrade but I would have to research more thoroughly how that might affect downstream users first.

miner46er:
I've rolled back the version to the last known good one.

LandonTClipp (approved):

LandonTClipp:
Thanks again! Merged.

PR #1061, opened 2025-06-02
Fix matryer template imports when outputting mock to different package than source

### LINKED ISSUE #1041: Go imports broken with v3.2.5
## Description

Go imports are broken with v3.2.5 (with v3.2.4 it is OK, so I assume, that https://github.com/vektra/mockery/pull/1032 did break this).

## Reproducer

I can not share the actual template I observed the issue with, but this stripped down template, produces the problem for me as well:

```none
package {{ .PkgName }}

import (
{{ range .Imports -}}
	{{ .Alias }} "{{ .Path }}"
{{ end -}}
)


{{ range $i, $wrapper := .Interfaces }}

{{ $decorator := (or .TemplateData.DecoratorName (printf "%sWithSlog" $wrapper.Name)) }}

type {{$decorator}} struct {
	_log  *slog.Logger
	_base {{ $.SrcPkgQualifier }}{{ $wrapper.Name }}
}

func New{{$decorator}}(base {{ $.SrcPkgQualifier }}{{ $wrapper.Name }}, log *slog.Logger) {{$decorator}} {
	return {{$decorator}}{
		_base: base,
		_log: log,
	}
}

{{ end }}
```

## Expected behavior

The above template is extracted from a wrapper (not a mock) around an interface using the `log/slog` package from the stdlib, but v3.2.5 fails to add the necessary import in this case. 
The import is not present in the interface, that is mocked, so mockery needs to figure out, what additional packages are used by the generated code. In the actual package, there is an other internal package, which also is used in the wrapper (but not referenced by the interface), which is also not added.
Interestingly, since I removed all the actual wrapping functions, `mockery` added the `context` package even though, the now generated code does no longer require it.

In my opinion, this is exactly what goimports is for and I don't share the opinion, that this should be handled by mockery it self. In this case I would really recommend to rely on the package provided by the Go team.

## Mockery version

v3.2.5

## Installation Mechanism

- [X] `go get`

## Go version

Go version: 1.23.9

chadwpetersen:
Description
-------------

This change updates the Matryer mock generation template to ensure the **source package of the interface** is correctly imported when generating a mock—particularly when the skip-ensure flag is not set.

When `skip-ensure` is **disabled**, the generated mock includes a **compiler assertion** to verify that the mock correctly implements the interface. However, if the interface being mocked resides in a **different package** from the generated mock, the compiler check fails unless the source interface's package is explicitly imported and referenced.

This update modifies the template logic to:
- Check whether the source and destination packages differ
- Inspect interfaces for the absence of the `skip-ensure` flag
- If needed, add the source package to the import list to support proper qualification of the interface in the compiler check

This ensures valid code generation even when mocking interfaces from external packages.
- Fixes #1041

## Type of change

- [ ] 1.22
- [ ] 1.23
- [x] 1.23

How Has This Been Tested?
---------------------------

- Manually generated mocks for interfaces located in a different package than the output mock file.
- Verified that:
  - The source package was added as an import if the skip-ensure flag was not set
  - The generated code compiled successfully
  - The compiler assertion `var _ pkg.SomeInterface = &MockType{}` referenced the interface with the correct package qualifier
- Regression tested against existing templates to ensure no impact when the interface is in the same package or skip-ensure is enabled

Checklist
-----------


internal/mock_matryer.templ
```diff
@@ -15,6 +15,15 @@ package {{.PkgName}}
     {{- $_ := .Registry.AddImport "sync" "sync" }}
 {{- end }}
 
+{{- if ne .PkgName .Registry.SrcPkg.Name }}
```
LandonTClipp: Is this correct in the general case? Technically, the mocks being placed outside of the source package path can still have the same package name, so I think this logic would incorrectly determine them to be in the same package in that case.

LandonTClipp (commented):
Thanks for the PR! Left a comment. Also, could you add a test to prove that this fixes the issue?

internal/mock_matryer.templ
chadwpetersen: Valid point. I changed this to `{{- if ne $.SrcPkgQualifier "" }}`. It seems like this value is set only when not in package. So we should only consider doing this if we are not in the source package. I have also added some tests that checks the three different cases -generating within package, generating outside of package with different name, and outside package with the same name as the source package.

internal/mock_matryer.templ
```diff
@@ -15,6 +15,15 @@ package {{.PkgName}}
     {{- $_ := .Registry.AddImport "sync" "sync" }}
 {{- end }}
 
+{{- if ne $.SrcPkgQualifier "" }}
+  {{- range .Interfaces }}
+    {{- if not (index .TemplateData "skip-ensure") }}
+      {{- $_ := $.Registry.AddImport $.SrcPkgQualifier $.Registry.SrcPkg.PkgPath }}
```
jhonndabi:
```suggestion
      {{- $_ := $.Registry.AddImport $.Registry.SrcPkgName $.Registry.SrcPkg.PkgPath }}
```

SrcPkgQualifier can have a `.` as suffix, and if there is a name collision, `AddImport` will not add an alias to resolve it, since import qualifiers will be different.

https://github.com/vektra/mockery/blob/v3.3.2/internal/template_generator.go?plain=1#L465

jhonndabi: Something like this should help solve this issue:

```golang
{{$srcPkgQualifier := $.SrcPkgQualifier}}
{{- if not (index .TemplateData "skip-ensure") }}
	{{- $_ := $.Registry.AddImport $.Registry.SrcPkgName $.Registry.SrcPkg.PkgPath }}
{{- end }}

import (
{{- range .Imports}}
	{{ .ImportStatement }}

	{{- if eq .Path $.Registry.SrcPkg.PkgPath  }}
		{{- $srcPkgQualifier = (printf "%s." .Qualifier) }}
	{{- end }}
{{- end}}
)
```
and in the code below, instead of using `{{$.SrcPkgQualifier}}` change it to `{{$srcPkgQualifier}}`

jhonndabi: This way it will be possible to use the alias attributed to `$.Registry.SrcPkg.PkgPath` in `AddImport`.

chadwpetersen: Ah yes, you have a valid point. I have updated the template as suggested here -> https://github.com/vektra/mockery/pull/1061/commits/591de4560d233f8de5fe5b58976724812e261c7a

LandonTClipp (approved):
This looks great, thanks again for the contribution!

Issue #1092, opened 2025-07-22
Mockery v3 Generates invalid code for private types


sonalys:
## Description

I have a package with a private interface and types
My configuration is set to generate the mocks on a different package name
Mockery will generate the mock for the private interface and reference a private type, which is invalid

## Expected behavior

Either:

* Configurable option to ignore private interfaces / types
* Mockery detects mocks are generated on a different package and ignores private interfaces / types

## Mockery version

`v3`

## Installation Mechanism

- [x] `go get`
- [] Pre-built release
- [] homebrew
- [] Other: [please describe]

LandonTClipp:
You can fix this by just excluding such interfaces that you know of using `exclude-interface-regex`. I know that's not ideal, but it would work. Your idea of having mockery detect such invalid cases is a good one, I'd be interested to hear more proposals on what this would look like in a concrete sense.

sonalys:
@LandonTClipp I pushed a draft. I'm not happy with some parts, like built-in detection, but it solves the issue for me.

It's not ready to be merged, I'm just showing you the problem.

If you want, we can iterate over it.

PR #1094, opened 2025-07-29
issue-1092: implement verification for private types

sonalys:
Description
-------------

Small draft for issue #1092 

Do not merge!

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

Added test files for confirming no generation of invalid code, while all other test fixtures remain unchanged.

Checklist
-----------


LandonTClipp:
Thank you! I'll take a look in a few days.

LandonTClipp:
I haven't yet looked too deeply into the implementation of this, but I think what we should do is retain the current behavior and add a warning log when mockery detects this that prompts the user to enable something like `disable-private-type-generation: true` if they want mockery to exclude such interfaces. The reason why is that there are some build pipelines of our users whereby they generate the mocks in a directory tree _outside_ of the original mocks but then copy that tree directly into the source code. In such cases, this PR would always exclude privately-typed mocks and break these pipelines.

We should also add a mechanism for users to disable the warning log. You might be able to borrow the logic in v2 that lets users disable certain deprecation warnings.

sonalys:
> I haven't yet looked too deeply into the implementation of this, but I think what we should do is retain the current behavior and add a warning log when mockery detects this that prompts the user to enable something like `disable-private-type-generation: true` if they want mockery to exclude such interfaces. The reason why is that there are some build pipelines of our users whereby they generate the mocks in a directory tree _outside_ of the original mocks but then copy that tree directly into the source code. In such cases, this PR would always exclude privately-typed mocks and break these pipelines.
> 
> We should also add a mechanism for users to disable the warning log. You might be able to borrow the logic in v2 that lets users disable certain deprecation warnings.

If the package name is different, even the scenario you mentioned is not applicable.

So maybe there is a need for 2 detection layers, a different package name, and a different package path.

PR #1105, opened 2025-08-16
Add support for overriding configuration through comments on interfaces

### LINKED ISSUE #1064: Support mock generation via interface Annotations
## Summary
I’d like to request support for generating mocks from interfaces explicitly annotated with a specific comment tag, such as `@mockery:generate`.

## Background
In my current workflow, I use `go:generate` to trigger mock generation for selected interfaces in a large repository. However, managing inclusion and exclusion through regex patterns has proven difficult due to:

- Not all interfaces require mocks.
- Interface names do not follow consistent patterns that are easily targeted via regex.

## Proposed Solution
Introduce an option in the `mockery.yml` configuration file:

```yaml
annotation: true
```

When `annotation: true` is set, mockery will only generate mocks for interfaces that include a specific comment annotation above them (e.g., `// @mockery:generate`). This could look like:

```go
// @mockery:generate
type MyInterface interface {
    ...
}
```

## Question

Is something like this currently possible with mockery, or perhaps already being planned? I couldn't find any mention of `annotation-based` generation in the documentation, so I’m wondering if there's an existing or alternative approach to achieve this use case.

paivagustavo:
Description
-------------

Adds support for configuring mocks through directive comments on the interface. The comment will be parsed as a yaml configuration, any configuration supported on the file based config is also supported by this:
```Go
//mockery:struct_name: <struct_name>
//mockery:file_name: <file_name>
//mockery:template: <template>
//mockery:<key>: <value>
```

These configuration will take precedene of over any other configuration. 

I haven't made any changes to documentation yet as I wanted to wait for a confirmation that this pr will be accepted before doing so.

- Fixes #1064

## Type of change

- [X] 1.24

How Has This Been Tested?
---------------------------

Added a new package with these annotations in the fixture folder, testing all the new configurations.

Checklist
-----------


config/interface_overrides.go
```diff
@@ -0,0 +1,83 @@
...
+
+		if matches := generateRe.FindStringSubmatch(text); len(matches) > 1 {
+			overrides.Generate = matches[1] == "true"
+		} else if matches := structNameRe.FindStringSubmatch(text); len(matches) > 1 {
+			overrides.StructName = strings.TrimSpace(matches[1])
+		} else if matches := fileNameRe.FindStringSubmatch(text); len(matches) > 1 {
+			overrides.FileName = strings.TrimSpace(matches[1])
+		} else if matches := templateRe.FindStringSubmatch(text); len(matches) > 1 {
+			overrides.Template = strings.TrimSpace(matches[1])
+		}
```
LandonTClipp: This is really good, but I think we can do even better. Obviously the problem with this way of doing it is that we must have manual support for each individual config parameter. Instead, I think we could design this in a way that it will support _all_ config parameters as defined in [`config.Config`](https://github.com/paivagustavo/mockery/blob/0f536ead4b49a1d5ebbf9269144178021a22fd3c/config/config.go#L523). 

The way you'd want to go about this is to parse the yaml in the doc comment into this config struct, then [merge this](https://github.com/paivagustavo/mockery/blob/0f536ead4b49a1d5ebbf9269144178021a22fd3c/config/config.go#L241) into the interface's `Config` struct. You already kind of do this below in your `Override` method, but it's cleaner to use the merge method already available.

Additionally, it'd be better if the override discovery logic lives [here](https://github.com/vektra/mockery/pull/1105/files#diff-2b1f49960c8524c032e7cd25400fa991b9e0c19911bb648aee4bf56f48e67d79R285) in `GetInterfaceConfig` instead of needing a separate call. I'd eventually like to bring the `showconfig` command into parity with what you've written here so that overrides are properly represented. Obviously you'd need to pass the `*ast.GenDecl` syntax to make that work, which is fine.

Overall great work, thank you for tackling this!

LandonTClipp (changes requested):

config/interface_overrides.go
```diff
@@ -0,0 +1,83 @@
...
+	"go/ast"
+	"regexp"
+	"strings"
+)
+
+var (
+	generateRe   = regexp.MustCompile(`mockery_generate: (true|false)`)
+	structNameRe = regexp.MustCompile(`mockery_struct_name: ([^\s]+)$`)
+	fileNameRe   = regexp.MustCompile(`mockery_file_name: ([^\s]+)$`)
+	templateRe   = regexp.MustCompile(`mockery_template: ([^\s]+)$`)
```
jamesphillpotts-fr: The [go doc language spec](https://tip.golang.org/doc/comment#syntax) talks about directives and states that

> Tools that define their own directives should use the form //toolname:directive.

It seems like the proposed directives should conform to this syntax too, so comments would be:

```golang
//mockery:<parameter name> <value>
```

If that makes sense?

LandonTClipp: That makes a lot of sense. I'm not sure if this PR is going to be updated, we might need someone else to take it over.

config/interface_overrides.go
paivagustavo: Sorry It took me so long for getting back to this! Yeah, I totally agree that this was a little bit too manual. I've refactor it a little bit to extract a yaml from the comment section and parse it into a config - merging it with the file based pkg/interface config if they exist.

Regarding the last piece:
> Additionally, it'd be better if the override discovery logic lives [here](https://github.com/vektra/mockery/pull/1105/files#diff-2b1f49960c8524c032e7cd25400fa991b9e0c19911bb648aee4bf56f48e67d79R285) in GetInterfaceConfig instead of needing a separate call.

If we were to do that, we wouldn't have loaded the config from the comments *before* the `ShouldGenerateInterface` check, this means that we would need to have the interface defined in the config file (unless we move this check to happen later).

config/interface_overrides.go
paivagustavo: Hey 👋  (sorry for being missing)!

I like that! I took a chance to refactor the initial pr with @LandonTClipp initial comment and this suggestion! Let me know what do you think.

internal/fixtures/directive_comments/directive_comments.go
```diff
@@ -0,0 +1,38 @@
+package directive_comments
+
+import "net/http"
+
+// Requester is an interface that defines a method for making HTTP requests.
+//
+//mockery:generate: true
```
LandonTClipp, line 7: My reading of the code is that the `generate: true` config is not actually doing anything other than serving as a trigger that says "hey, there is a `mockery:` directive here". We could have put anything here, like `//mockery:foo: bar` and the same thing would happen. Is that accurate? If so, the `generate:` config should have a real meaning, as in, the presence of this config (and not others) triggers the mock to be generated. Furthermore, I think this should really just be `//mockery:generate` to conform with the same standard as `//go:generate`.

LandonTClipp (commented):
This is much, much better! I need to sit with this and ponder the implications of this PR a bit more. I might submit more comments in the coming days. Overall, this is probably close to being done.

config/config.go
```diff
@@ -695,3 +735,41 @@ func (c *Config) GetReplacement(pkgPath string, typeName string) *ReplaceType {
...
+
+	var yamlConfig []string
+
+	// Extract all mockery directive comments and build a YAML document
+	for _, doc := range decl.Doc.List {
+		// Look for directive comments  `//mockery:<config-key>: <value>` and convert them to YAML
+		if value, found := strings.CutPrefix(doc.Text, "//mockery:"); found && value != "" {
+			yamlConfig = append(yamlConfig, value)
+		}
+	}
```
jamesphillpotts-fr, line 766: I can see that this is an easy and convenient way to translate into mockery config, but it might be nice to have some alternative syntax, rather than spreading yaml through multiple directive lines, especially if you could take advantage of the provided [`ast.ParseDirective`](https://pkg.go.dev/go/ast@master#ParseDirective) function?

Something like this might be nicer, but I appreciate it may be a bit more work...

```golang
//mockery:generate -structname MyMock -filename my_mock.go
```

LandonTClipp: I err on the side of being obvious versus being clever. Emulating command line args with mockery actually having any command line args is confusing and weird. I'm okay with the way this is currently implemented.

LandonTClipp:
@paivagustavo I have a strange request, can you give me write permission on your fork? I have some tweaks I want to add to your branch regarding the `ShouldGenerateInterface` logic. I'm excited about this PR but I just wanted to collaborate a little more directly.

paivagustavo:
> @paivagustavo I have a strange request, can you give me write permission on your fork? I have some tweaks I want to add to your branch regarding the `ShouldGenerateInterface` logic. I'm excited about this PR but I just wanted to collaborate a little more directly.

Of course @LandonTClipp, I've added you! feel free to do any change :)

LandonTClipp (approved):

paivagustavo:
@LandonTClipp your changes looks good to me! I've added one [last commit](https://github.com/vektra/mockery/pull/1105/commits/8e4967c4e9ae041abb39b7dafb97dfbf51c2cd3a) to remove a println that I've added during some tests :). Thank you for having patience with me!

LandonTClipp:
No worries, I think we're ready to merge this now once the tests pass. I bumped the minor version so it will get released automatically. Thanks again, this is great!

jamesphillpotts-fr:
Nice change, thanks @paivagustavo @LandonTClipp! :tada:

PR #1125, opened 2025-11-13
fix: constraint interfaces generation

### LINKED ISSUE #1113: Mocks generated for constraint interfaces/anys
## Description

Upon upgrading from `v3.3.4` to `v3.5.4`, we found that mockery is now generating mocks for some categories of types that I don't think it should:

- `type A constraints.Ordered`
- `type B any`

I also realised that mockery was already generating mocks for:
- `type C interface { constraints.Float | constraints.Integer }`

## Reproducer

Create types like the ones above and run mockery

## Expected behavior

Mocks shouldn't be generated for those types

## Mockery version

`v3.5.4`

## Installation Mechanism

- [x] `go get`
- [ ] Pre-built release
- [ ] homebrew
- [ ] Other: [please describe]

## Go version

v1.25

Alhanaqtah:
Description
-------------

I added a check to the parser for whether the interface is a constraint. Mocks will not be generated using such interfaces.

Examples of interfaces that will be skipped:
```go
type Skip1 constraints.Ordered

type Skip2 interface {
	~int
}

type Skip3 interface {
	constraints.Float
}

type Skip4 interface {
	constraints.Float | constraints.Integer
}

type Skip5 Skip1

type Skip6 interface {
	Skip5
}
```
- Fixes #1113

## Type of change

- [x] 1.25.1

How Has This Been Tested?
---------------------------
I wrote a test in fixtures with several constraint interfaces. If no mocks are generated for them, the test will be completed successfully.
Fixture path: ```internal/fixtures/constraint_ifaces```

Checklist
-----------


internal/parse.go
```diff
@@ -171,3 +176,45 @@ func isAutoGenerated(pathName string) (bool, error) {
...
+			// Example:
+			//	type I interface {
+			//		constraints.Float
+			//	}
+			if sel, ok := item.Type.(*ast.SelectorExpr); ok {
+				if id, ok := sel.X.(*ast.Ident); ok && id.Name == "constraints" {
+					return true
+				}
+			}
+		}
```
LandonTClipp: I think this is fine but it's easily tricked. You can create an alias to a `constraints.Float` and this check would not catch it. You would need an additional check for alias types, then resolve the alias, then perform the same checks you added here.

That doesn't _have_ to be included in this PR, but it's something to think about. You mention this in the description so I'm not telling you something new.

Alhanaqtah: Hi. Thanks for the tip. Thanks to this, i also found a way to check aliases for constraint interfaces

LandonTClipp:
I haven't forgotten this, I will get to it soon!

LandonTClipp (approved):

LandonTClipp:
Merged, thanks again 🫡

PR #1138, opened 2026-01-21
feat: goimports configuration options

### LINKED ISSUE #1098: Feature request: support formatter options like goimports local prefix
I'm currently upgrading a project to mockery v3 and am setting up a config from scratch. We're using goimports with the local prefix option to separate public modules from private ones in our imports. So now I need to wrap a mockery call in a make command to always run goimports with the local prefix option afterwards. It's not a huge deal, but I'd love to be able to set the local prefix option when configuring mockery to use goimports directly.

I'd be happy to contribute an implementation proposal, but I wanted to clarify if you would even consider adding formatter options to the mockery config.

In terms of configuration structure, I think something like this could make sense:

```yaml
...
formatter: goimports
formatter-options:
  goimports:
    local-prefix: example.com/my-org
```

rfwatson:
Description
-------------

Add goimports local-prefix configuration. Fixes #1098.

## Type of change

- [ ] 1.22
- [x] 1.23

How Has This Been Tested?
---------------------------

Mostly manual testing. It seems that testing the actual formatter call could be a bit tricky, happy to take a look at adding some if you think it's worth it though.

Checklist
-----------


config/config.go
```diff
@@ -571,16 +577,17 @@ type Config struct {
...
-	IncludeAutoGenerated  *bool   `koanf:"include-auto-generated" yaml:"include-auto-generated,omitempty"`
-	IncludeInterfaceRegex *string `koanf:"include-interface-regex" yaml:"include-interface-regex,omitempty"`
-	InPackage             *bool   `koanf:"inpackage" yaml:"inpackage,omitempty"`
-	LogLevel              *string `koanf:"log-level" yaml:"log-level,omitempty"`
-	StructName            *string `koanf:"structname" yaml:"structname,omitempty"`
-	PkgName               *string `koanf:"pkgname" yaml:"pkgname,omitempty"`
-	Recursive             *bool   `koanf:"recursive" yaml:"recursive,omitempty"`
+	ForceFileWrite        *bool            `koanf:"force-file-write" yaml:"force-file-write,omitempty"`
+	Formatter             *string          `koanf:"formatter" yaml:"formatter,omitempty"`
+	FormatterOptions      FormatterOptions `koanf:"formatter-options" yaml:"formatter-options,omitempty"`
```
LandonTClipp: This should be `*FormatterOptions` so that the yaml map can be nil.

internal/template_generator.go
LandonTClipp: This is good but I think people also want to be able to modify the `imports.Options` struct. I think we can do a functional option style of thing. I suggest something like this:

```go
type FormatterOptions struct {
	GoImports *GoImports
}

type GoImports struct {
	LocalPrefix *string ...
    TabWidth *int ...
}

func (g GoImports) Options() imports.Options {
    return imports.Options{
      TabWidth: g.TabWidth,
      // etc
    }
}
```

Then:

```go
func goimports(src []byte, opt GoImports) ([]byte, error) {
    imports.LocalPrefix = opt.LocalPrefix
    formatted, err := imports.Process("/", src, opt.Options())
```

We'd want to set default values in here: https://github.com/rfwatson/mockery/blob/dc7ca2f8c944d58757e85aef489fee65a5da13fd/config/config.go#L68-L85

LandonTClipp (commented):
Looks great, let's just extend this a bit more.

config/config.go
rfwatson: Wouldn't we just end up with a zero value `FormatterOptions` in that case? I was thinking to do it this way to avoid the nil checks later, might be missing something though.

LandonTClipp: The reason is pretty complicated. It has to do with the migrate command when going from v2 to v3. If you set the default in koanf then mockery will never have nil values for it.

internal/template_generator.go
rfwatson: Updated

config/config_test.go
```diff
@@ -73,6 +73,60 @@ packages:
...
+	require.NotNil(t, cfg.FormatterOptions.GoImports)
+	assert.Equal(t, "goimports", *cfg.Formatter)
+	assert.True(t, *cfg.FormatterOptions.GoImports.AllErrors)
+	assert.False(t, *cfg.FormatterOptions.GoImports.Comments)
+	assert.True(t, *cfg.FormatterOptions.GoImports.FormatOnly)
+	assert.True(t, *cfg.FormatterOptions.GoImports.Fragment)
+	assert.Equal(t, "github.com/vektra/mockery", *cfg.FormatterOptions.GoImports.LocalPrefix)
+	assert.False(t, *cfg.FormatterOptions.GoImports.TabIndent)
+	assert.Equal(t, 4, *cfg.FormatterOptions.GoImports.TabWidth)
+}
```
LandonTClipp: Testing that mockery can unmarshal values isn't really a useful thing to test. We can probably remove this.

LandonTClipp (changes requested):
Just one change, we can delete one of the tests. Once that's done, let's get it merged!

config/config_test.go
rfwatson: Removed.

I think the original intention was to test that nil values were handled correctly. Everything got refactored a couple of times since writing it though, so it's quite possibly not as relevant as it was.

LandonTClipp (approved):
This is great! I'm going to give it one last manual verification before merging.

LandonTClipp:
Actually, it looks like the tests are revealing backwards-incompatible behavior: https://github.com/vektra/mockery/pull/1138/changes#diff-dd1c701f26e4b66f2c1c1b312fd78581f706770353861da4095b1fbf740c47deL490-L494

These are the values mockery provided by default so we should keep `formatonly` to true.

rfwatson:
Updated.

Which test is failing though? I only see [these](https://github.com/vektra/mockery/actions/runs/22592933573/job/65462238212) which don't look related?

LandonTClipp:
The test was showing that some mock generation was producing a different result than what was committed (git diff). You fixed the issue but now there is a test that needs its default value updated. Then, we should be good to go.

LandonTClipp:
Wonderful, thank you @rfwatson for the contribution!

Issue #1019, opened 2025-04-18
v3: package name collision problem


spirin:
When regenerating mocks, a name conflict may occur for packages with a similar name.

`force-file-write: true ` does not help.

I have prepared a simplified project that shows the problem.
[mockery-test.zip](https://github.com/user-attachments/files/19811627/mockery-test.zip)

Steps to reproduce:
1. open project mockery-test
1. run  `mockery`
1. run  `mockery` again

The key factor here is that the package "mockery-test/uuid" is not called `uuid` but `package another_uuid`

<img width="428" alt="Image" src="https://github.com/user-attachments/assets/18905a5d-1da9-4d6e-ba71-ae8b490d8957" />

LandonTClipp:
Interesting, it appears something about one of these packages is not being registered with the naming scope which means it's not doing package name aliasing properly. I'll have to take a look.

LandonTClipp:
Can you confirm what version of mockery you're running? When I run on v3.2.3, I'm able to generate a valid mock file and it handles the naming collision properly. Example:

```go
// Code generated by mockery; DO NOT EDIT.
// github.com/vektra/mockery
// template: testify
// TEST MOCKERY BOILERPLATE

package main

import (
	"github.com/google/uuid"
	mock "github.com/stretchr/testify/mock"
	uuid0 "github.com/vektra/mockery/v3/internal/fixtures/package_name_collision/uuid"
)
```

I had to change the package name of `github.com/vektra/mockery/v3/internal/fixtures/package_name_collision/uuid` to `uuid` instead of `another_uuid` to force the collision to happen, and you can see mockery properly changed it to `uuid0`. You can see an example in the linked commit above.

spirin:
v3.2.2, v3.2.3.

Just repeat without modifications my steps to reproduce that I described

> I had to change the package name

no, it should be `another_uuid` to reproduce. And you need to run it TWO times.

config:
```yaml
all: true
recursive: true
filename: mocks_generated.go
force-file-write: true
packages:
  mockery-test:

```

[Screen Recording 2025-04-23 at 08.59.30.zip](https://github.com/user-attachments/files/19861724/Screen.Recording.2025-04-23.at.08.59.30.zip)

LandonTClipp:
I see, I'm able to reproduce it internally as well as your example. I will take a closer look.

LandonTClipp:
[...]

Issue #935, opened 2025-03-03
Findings while trying v3 for matryer/moq style mocks


breml:
Findings:

* In the config, the accepted value for `template` is `matryer`, but the [documentation](https://vektra.github.io/mockery/v3/configuration/#parameter-descriptions) does mention `moq`.
* In the generated mock, it still says: `If this is not the case, regenerate this file with moq.`, I think this should be changed to `If this is not the case, regenerate this file with mockery.`. Location in the template: https://github.com/vektra/mockery/blob/v3/internal/mock_matryer.templ#L27
* With `moq`, if the mocked interface has used a type from an other package, that has been imported using an alias, this alias has also been used in the generated mock. This is no longer the case with mockery.

breml:
One more thing, I just saw, that for `beta.1` the toolchain requirement is Go 1.24. Is there a technical reason for this? In corporate environments, there is often a requirement to stick with `oldstable` until a new Go version is released. Therefore I would appreciate, if mockery V3 would only require `oldstable` (Go 1.23 as of now) as a hard requirement.

LandonTClipp:
> One more thing, I just saw, that for beta.1 the toolchain requirement is Go 1.24. Is there a technical reason for this? In corporate environments, there is often a requirement to stick with oldstable until a new Go version is released. Therefore I would appreciate, if mockery V3 would only require oldstable (Go 1.23 as of now) as a hard requirement.

Go 1.24 was required because of the introduction of generic type aliases, so we need an updated toolchain to support parsing this. If this existing in the toolchain is an issue, there might be a more targeted way we can go about this. We need an updated parser (from `golang.org/x/tools/go/packages`) and AST support. If there's a better way to do this, I'm happy to look into it.

breml:
> Regarding:
> 
> ```
> With moq, if the mocked interface has used a type from an other package, that has been imported using an alias, this alias has also been used in the generated mock. This is no longer the case with mockery.
> ```
> 
> I'm curious to know if there is a practical reason why this is an issue? This is pretty much expected, as mockery will dynamically decide whether or not to use an alias for an import, which is essentially just a function of what other names are visible in the file-global scope.

As far as I can tell until now, I have not yet hit a case, where this became an issue, so I mainly raised this because of the additional changed lines, this caused in the diff of the PR, where I would have wished, that the changes are minimal if `mockery` is used as a "drop-in-replacement".
[...]

Issue #621, opened 2023-05-12
Mockery v2 with `--inpackage` generates import path that conflicts with type name, ignoring aliased import in source file


geekofalltrades:
Description
-------------

I have the following code:

```go
package myclient

import (
	awsclient "github.com/aws/aws-sdk-go/aws/client"
	"github.com/aws/aws-sdk-go/service/ecr"
	"github.com/aws/aws-sdk-go/service/ecr/ecriface"
)

// ecrAPIExtension is an extension of ecriface.ECRAPI that adds getter methods
// for useful exported attributes of ECRAPI implementations.
type ecrAPIExtension interface {
	ecriface.ECRAPI

	GetClient() *awsclient.Client
}

// ecrClient is an implementation of ecrAPIWithClient for production use.
type ecrClient struct {
	*ecr.ECR
}

func (e *ecrClient) GetClient() *awsclient.Client {
	return e.Client
}

// client implements containerregistry.Client to manage ECR.
type client struct {
	ecr ecrAPIExtension
}
```

It defines a `client` type that's implementing an interface defined elsewhere. I am also importing `awsclient "github.com/aws/aws-sdk-go/aws/client"`. It's being aliased to `awsclient` to avoid a name clash with the `client` type.

I run the following mockery command to generate a mock of the `ecrAPIExtension` type:

```
mockery --dir path/to/mypackage/ --all --log-level debug --inpackage
```

The resulting mock has all of the methods of the embedded [`ecriface.ECRAPI`](https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/ecriface/#ECRAPI), so it's very large, and I won't paste the whole file here, but here's the relevant portion:

```go
// Code generated by mockery v2.26.1. DO NOT EDIT.

package containerregistry

import (
	context "context"

	client "github.com/aws/aws-sdk-go/aws/client"

	ecr "github.com/aws/aws-sdk-go/service/ecr"

	mock "github.com/stretchr/testify/mock"

	request "github.com/aws/aws-sdk-go/aws/request"
)

// mockEcrAPIExtension is an autogenerated mock type for the ecrAPIExtension type
type mockEcrAPIExtension struct {
	mock.Mock
}

# etc.
```

It has ignored the `awclient` alias I imported `"github.com/aws/aws-sdk-go/aws/client"` with and is importing it as `client`, instead. This conflicts with my `client` type in the same package.

Mockery Version
--------

v2.26.1

Golang Version
--------------

[...]

PR #1170, opened 2026-08-02
v3: avoid in-package import collisions with package-scope declarations

dgollahon-plaid:
## Summary

- Reserve source package-scope names when generating an in-package mock.
- Alias imports that collide with those names using mockery's existing numbered fallback.
- Add focused registry tests and a compile-enforcing generated fixture.

## Problem

Mockery v3 can generate an unaliased import that conflicts with a declaration in the destination package. The generated code does not compile.

Minimal source package:

```go
package dep

import libdep "example.com/repro/lib/dep"

type dep struct{}

type Iface interface {
	M() libdep.T
}
```

With `inpackage: true`, v3 generated this import:

```go
import "example.com/repro/lib/dep"
```

The package then failed to build:

```text
dep already declared through import of package dep ("example.com/repro/lib/dep")
    other declaration of dep
```

| Version | Result |
| --- | --- |
| v2.53.6 | Preserves a non-conflicting alias and compiles |
| v3.6.4 | Generates the bare `dep` import and fails to compile |
| v3.7.1 | Generates the bare `dep` import and fails to compile |
| v3 HEAD `88e7c13` | Generates the bare `dep` import and fails to compile |
| This PR | Generates `dep0` and compiles |

V3 has no configuration option that can avoid this collision.

## Root cause and fix

`Registry.addImport` checked candidate qualifiers only against imports already selected for the generated file. For in-package output, package-scope declarations are also visible and can use the same name.

`NewRegistry` now records names from the source package's type scope when the output is in-package. `addImport` checks that reserved set together with existing import qualifiers. This keeps the existing import-to-import behavior and numbered alias selection unchanged.

This does not attempt to preserve a source file's human-written import alias. That can be handled as a separate enhancement.

## Tests

- Added registry tests for an in-package collision, out-of-package behavior, and an occupied numbered alias.
- Added an in-package fixture whose generated mock imports a package named `dep` alongside a package-scope type named `dep`.
- Regenerated testify and matryer fixtures with no changes to existing generated files.
- `go test ./...`
- Standalone reproduction: unpatched HEAD fails with the compiler error above; this branch emits `dep0` and `go build ./...` passes.

## Related issues

- #935 finding 3 reports that v3 no longer preserves import aliases. This case shows that the behavior can be a correctness issue, not only generated diff noise.
- #621 describes the equivalent in-package compile collision and the expected v2 behavior.
- #1019 covers a different collision between a directory name and a declared package name.


LandonTClipp:
This is a great fix, I have no suggestions. Thank you!

LandonTClipp (approved):

Issue #914, opened 2025-02-11
Go 1.24 compatibility / `package "context" without types was imported` error


mikeauclair:
Seeing an error generating mocks with Go 1.24.0
```
2025/02/11 20:30:02 internal error: package "context" without types was imported from "<snipped>/client"
```
where the import is more-or-less just
```
package client

import (
  "context"
)
```

It looks like this has been a common issue in the past with new Go versions, and running through the cache clearing steps referred to in https://vektra.github.io/mockery/v2.35/notes/#internal-error-package-without-types-was-imported didn't appear to help so I figured I'd flag this in case it's an actual issue w/ 1.24.0

Thanks!

LandonTClipp:
Thanks for the report. I'll have to look into this, usually these problems are environmental and involve stale cached packages, but since you said you already cleaned the cache and are still encountering the problems, there might be something novel going on. 

Can you provide more details on your running environment:

1. 
    ``` 
    $ go version
    go version go1.23.5 darwin/arm64
    ```
2. `uname -a`

This project also does not yet official support 1.24 as you can see the go.mod file is at 1.23 and 1.24 is not in the testing matrix yet. I just need to spend some time to bringing it up to spec.

firefart:
Sure

```
go version
go version go1.24.0 linux/amd64
uname -a
Linux zoidberg 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

Failure is also reproducible on github actions when running mockery

mikeauclair:
As requested:
```
$ go version
go version go1.24.0 linux/amd64
$ uname -a
Linux <hostname> 6.8.0-1021-aws #23-Ubuntu SMP Mon Dec  9 23:59:34 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

(also occurring in our self-hosted GHA runners on arm64 / debian 22.04)

> This project also does not yet official support 1.24 as you can see the go.mod file is at 1.23 and 1.24 is not in the testing matrix yet. I just need to spend some time to bringing it up to spec.

Totally understood, and thanks!

LandonTClipp:
I reproduced the problem in CI: https://github.com/vektra/mockery/actions/runs/13278535376/job/37072470870?pr=915

[...]

Issue #535, opened 2023-02-01
go 1.20 internal error: package "..." without types was imported from "..."


dillonstreator:
Description
-------------

```sh
01 Feb 23 16:35 CST INF Starting mockery dry-run=false version=v2.16.0
01 Feb 23 16:35 CST INF Walking dry-run=false version=v2.16.0
2023/02/01 16:35:53 internal error: package "context" without types was imported from "..."
```

Mockery Version
--------

v2.16.0

Golang Version
--------------

1.20

Installation Method
-------------------

- [ ] Binary Distribution
- [ ] Docker
- [X] brew
- [ ] go install
- [ ] Other: [specify]

Steps to Reproduce
------------------

generating mocks for the following package definition results in error

```go
package xyz

import "context"

type Thing interface {
	Do(ctx context.Context)
}

type Something struct{}

var _ Thing = (*Something)(nil)

func (s *Something) Do(ctx context.Context) {
}
```

dillonstreator:
bumping to `golang.org/x/tools@v0.5.0` should resolve

Issue #475, opened 2022-06-08
Running under go 1.18 throws 'package without types' error


pantsmann:
Description
-------------

`go generate` runs `mockery --name .*`
When I ran this under go 1.17 there are no problems. After upgrading to go 1.18 I get an error: 
`internal error: package "privaterepo/pkgone" without types was imported from "privaterepo/pkgtwo"`

Version
--------

`mockery --version` outputs:
`07 Jun 22 18:48 MDT INF Starting mockery dry-run=false version=v0.0.0-dev
v0.0.0-dev`

I installed mockery_2.13.0-beta.1_Darwin_x86_64.tar.gz

Steps to Reproduce
------------------

Explained above. Note: I kept go 1.17 on my machine and running `go1.17 generate` has no problem.

### Expected Behavior

mocks to be created

### Actual Behavior

Got an error.  See description

LandonTClipp:
Hi, can you provide a reproducer?

bandesz:
We had the same issue using Go 1.18.2, but after upgrading mockery to v2.13.0, the errors went away.

AlexCuse:
I was seeing this error with an older version as well but it stopped on update.  I am getting invalid mocks generated now (imports "github.com/nats-io/") but I think it is probably due to weirdness of my interface.  

```
import "github.com/nats-io/nats.go"

type Msg = nats.Msg
type MsgHandler = nats.MsgHandler
type Subscription = nats.Subscription

// Connection provides an interface over basic *nats.Conn methods that we need to interact with the broker
type Connection interface {
	// QueueSubscribe subscribes to a NATS subject, equivalent to default Subscribe if queuegroup not supplied.
	QueueSubscribe(string, string, MsgHandler) (*Subscription, error)
	// PublishMsg sends the provided NATS message to the broker.
	PublishMsg(*Msg) error
	// Drain will end all active subscription interest and attempt to wait for in-flight messages to process before closing.
	Drain() error
}
```

maja42:
This happens to me since I updated to go1.20

jonmol:
Having the same problem with go 1.20.1. Created a simple repo for reproduction https://github.com/jonmol/mockery-test

LandonTClipp:
I ran your reproducer and was not able to get the same error:

```
$ go generate generate.go
22 Mar 23 10:16 CDT INF Starting mockery dry-run=false version=v2.20.0
[...]

PR #445, opened 2022-04-11
fix: bump golang.org/x/tools to v0.1.10 to fix some go 1.18 issues

MathiasGr:
Description
-------------

Bump the [golang.org/x/tools](https://pkg.go.dev/golang.org/x/tools@v0.1.10) dependency to v0.1.10 to fix some issues during code gen.

Why bumping x/tools specifically? I got a hint from the ent/ent repo (see [issue](https://github.com/ent/ent/issues/2155), [PR](https://github.com/ent/contrib/pull/276/files)) that had the exact same issue recently and fixed it by doing so.

Example running mockery with go 1.18 locally before the fix:

```sh
mockery <...>
config dir: <...>
config output:<...>
2022/04/11 20:11:11 internal error: package "context" without types was imported from "<...>"
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

Version of Golang used when building/testing:
---------------------------------------------

- [ ] 1.11
- [ ] 1.12
- [ ] 1.13
- [ ] 1.14
- [ ] 1.15
- [ ] 1.16
- [x] 1.17
- [x] 1.18

How Has This Been Tested?
---------------------------

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Checklist
-----------

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1172, opened 2026-08-20
fix: bump golang.org/x/tools to v0.49.0 for Go 1.27 support

### LINKED ISSUE #1171: Go 1.27 compatibility: "internal error: package X without types was imported" (needs golang.org/x/tools bump for pkgbits V3/V4)
## Description

Running mockery (v3.7.1, and confirmed still present on the latest v3.7.3 release and on the unreleased `master` branch) under **Go 1.27.0** fails during package loading with:

```
internal error: package "fmt" without types was imported from "<module>/internal/app/.../rules"
```

This happens before any mocks are generated — mockery fails at the "Parsing configured packages..." stage.

## Root cause

Go 1.27 migrated the compiler's type export data format to `pkgbits` V3/V4 (tracked in [golang/go#78242](https://github.com/golang/go/issues/78242)). Tools built against an older `golang.org/x/tools` that doesn't understand this format fail with the classic `"internal error: package X without types was imported"` message when they try to read export data produced by a Go 1.27 compiler via `golang.org/x/tools/go/packages`.

mockery pins `golang.org/x/tools v0.42.0` in its `go.mod` (both `v3` and the `v2`-line `master` branch), which does not yet support this new format.

This is the same category of issue previously fixed for Go 1.18/1.20/1.24 in #475, #535, #914, and PR #445 — each time the fix was bumping `golang.org/x/tools`.

## Steps to reproduce

1. Use Go 1.27.0 (`go install`'d or via `actions/setup-go`).
2. In a module with a `.mockery.yaml` config, run:
   ```
   go install github.com/vektra/mockery/v3@v3.7.3
   mockery
   ```
3. Observe:
   ```
   INF Starting mockery config-file=.../.mockery.yaml version=v3.7.3
   INF Parsing configured packages...
   internal error: package "fmt" without types was imported from "..."
   ```

## Expected behavior

mockery should generate mocks successfully under Go 1.27.0, the same as it does under Go 1.26.x.

## Environment

- mockery: v3.7.1 and v3.7.3 (both affected); `golang.org/x/tools` pinned at `v0.42.0` in both
- Go: 1.27.0
- OS: Linux (GitHub Actions `ubuntu-24.04`), also expected to reproduce on other platforms

## Suggested fix

Bump `golang.org/x/tools` in `go.mod` to a version that supports reading Go 1.27's `pkgbits` V3/V4 export data (latest `golang.org/x/tools` as of this writing is v0.49.0).

I searched existing issues/PRs and didn't find one already tracking Go 1.27 support — apologies if this is a duplicate.

ergousha:
Fixes #1171

## Problem

mockery fails under **Go 1.27** while parsing configured packages:

```
INF Starting mockery config-file=.../.mockery.yaml version=v3.7.3
INF Parsing configured packages...
internal error: package "fmt" without types was imported from "<module>/internal/app/.../rules"
```

Go 1.27 emits type export data in the new `pkgbits` V3/V4 format ([golang/go#78242](https://github.com/golang/go/issues/78242)). `golang.org/x/tools v0.42.0` — the version currently pinned in `go.mod` — cannot read that format, so `go/packages` hands back packages without type information and mockery aborts.

This is the same class of breakage previously fixed for Go 1.18/1.20/1.24 (#475, #535, #914, #445), each time by bumping `golang.org/x/tools`.

## Fix

Bump `golang.org/x/tools` from `v0.42.0` to `v0.49.0` (`go get golang.org/x/tools@v0.49.0`). This also carries the usual transitive bumps of `x/mod`, `x/sync` and `x/sys`.

Only `go.mod` / `go.sum` change — no source changes, and **no change to the `go` directive** (still `go 1.25.5`).

## Verification

I bisected the failure to the `x/tools` version by comparing the dependency actually linked into the binary:

| build | `golang.org/x/tools` | result under Go 1.27.0 |
| --- | --- | --- |
| `go install github.com/vektra/mockery/v3@v3.7.1` | v0.42.0 | ❌ `internal error: package "fmt" without types...` |
| `go install github.com/vektra/mockery/v3@v3.7.3` | v0.42.0 | ❌ same failure |
| build inside this repo's `go.work` workspace | v0.48.0 | ✅ works |
| **this PR** (module mode, no workspace) | **v0.49.0** | ✅ works, exit code 0 |

Worth calling out: building inside the repo's `go.work` workspace masks the bug, because the workspace resolves `x/tools` up to v0.48.0. Users installing via `go install` get the `go.mod` pin (v0.42.0) and therefore hit the failure — which is why this isn't caught by the current CI setup.

Additional checks:

- `go build ./...` passes with `GOTOOLCHAIN=go1.25.5` (the `go.mod` minimum), so the bump does not raise the minimum supported Go version.
- `go test ./...` passes — 41/41 packages ok, 0 failures — both with the workspace active and with `GOWORK=off`, under Go 1.27.0.
- Verified end-to-end against a real project that previously failed: mockery now generates all mocks and exits 0.

## Note on CI

Since the workspace build hides this problem, it may be worth also adding a Go 1.27 job that exercises a `go install`-style (non-workspace) binary, so future export-format changes get caught. Happy to add that here if you'd like.


BrunoKrugel:
bump!

LandonTClipp:
Wow I take a break for 3 days and get one of the most popular PRs in this project's history.

Wtf did they put in 1.27?

LandonTClipp:
> Since the workspace build hides this problem, it may be worth also adding a Go 1.27 job that exercises a go install-style (non-workspace) binary, so future export-format changes get caught. Happy to add that here if you'd like.

If you could add this as a separate PR, that would be very much appreciated. Thank you!

PR #1178, opened 2026-08-24
fix: report unparsable `exclude-subpkg-regex` instead of panicking

### LINKED ISSUE #1177: Unparsable `exclude-subpkg-regex` panics instead of reporting a config error
## Description

An unparsable `exclude-subpkg-regex` pattern crashes mockery with a Go stack trace instead of a config error. `Config.ShouldExcludeSubpkg` (`config/config.go:674`) calls `panic(err)` on the `regexp.MatchString` error. `ShouldGenerateInterface` in the same file handles the same class of user input differently: an unparsable `include-interface-regex` or `exclude-interface-regex` is wrapped and returned, so it reaches the user as ``FTL app failed error="evaluating `include-interface-regex`: ..."``.

The patterns are not compiled or validated anywhere before this point, so any typo in `exclude-subpkg-regex` reaches `panic` whenever `recursive: true` is also set.

## Reproducer

```
$ cat .mockery.yaml
all: true
packages:
  example.com/demo:
    config:
      recursive: true
      exclude-subpkg-regex:
        - "("

$ mockery
panic: error parsing regexp: missing closing ): `(`

goroutine 1 [running]:
github.com/vektra/mockery/v3/config.(*Config).ShouldExcludeSubpkg(...)
	config/config.go:674 +0xa4
github.com/vektra/mockery/v3/config.(*RootConfig).Initialize(...)
	config/config.go:364 +0x99c
github.com/vektra/mockery/v3/config.NewRootConfig(...)
	config/config.go:241 +0x8f4
...
```

Same directory, same typo moved to a sibling option:

```
$ cat .mockery.yaml
packages:
  example.com/demo:
    config:
      include-interface-regex: "("

$ mockery
... FTL app failed error="evaluating `include-interface-regex`: error parsing regexp: missing closing ): `(`"
```

## Expected behavior

`exclude-subpkg-regex` reports a config error the way the other two regex options do.

## Mockery version

`v3.7.5-0.20260823053242-ffa1c3f8d526` (built from `v3` at ffa1c3f8)

## Installation Mechanism

- [x] Other: built from source at `v3` HEAD

## Go version

`go1.26.3 darwin/arm64`

## Note on the fix

Two remedies look reasonable: return the error from `ShouldExcludeSubpkg`, or compile the patterns during config validation. The first matches `ShouldGenerateInterface`, whose only caller already returns an error. Happy to send a PR on that shape if you agree.

One thing found while writing the regression test: the `test` task runs `./internal/... ./template_funcs/... ./template/...`, so `config/config_test.go` and its 7 test functions have not been executed by CI since b847fc8f moved the package out of `template/`.

dylanpulver:
Description
-------------

- Fixes #1177

`ShouldExcludeSubpkg` has one caller, inside a function that already returns `error`, so the failure has a path to the user without new plumbing.

Rejected alternative: compiling the patterns during config validation. That leaves the `panic` reachable on any other call path and splits the pattern lifecycle across two places. Returning the error matches `ShouldGenerateInterface`, which wraps the identical `regexp.MatchString` failure for `include-interface-regex`.

Residual: `ShouldExcludeSubpkg` is exported and its signature changes. The package doc states the package is not meant for external import and does not guarantee backwards compatibility of its methods.

On the `Taskfile.yml` line: the `test` task ran `./internal/... ./template_funcs/... ./template/...`, so `config/config_test.go` and its 7 existing test functions stopped running when b847fc8f moved the package from `template/` into `config/`. Adding `./config/...` is what makes the regression test execute in CI. Splitting it out is fine if you prefer.

## Type of change

Neither type box is ticked because the change fixes a bug but alters an exported signature. Your call on which applies given the package doc.

- [ ] 1.22
- [ ] 1.23

Neither. Built and tested with `go1.26.3 darwin/arm64`.

How Has This Been Tested?
---------------------------

Base ffa1c3f8, golangci-lint v2.13.1 from `tools/go.mod`.

- Lint: 0 issues on a clean checkout and 0 after.
- `GOWORK=off go test -count=1 ./config/... ./internal/... ./template_funcs/... ./template/...` and `./e2e/...` pass. `bash ./e2e/run_all.sh` then `git status` leaves the tree clean.
- Anti-vacuity: restoring `panic(err)` under the new signature fails `TestShouldExcludeSubpkg/unparsable_pattern` and `TestNewRootConfigUnparsableExcludeSubpkgRegex`. The other three table cases pin current behaviour rather than add coverage.
- Reproducer from #1177 now prints ``FTL app failed error="initializing root config: evaluating `exclude-subpkg-regex`: error parsing regexp: missing closing ): `(`"`` and exits nonzero.

Checklist
-----------

The unticked boxes are review judgements rather than measurements, so they are left for you.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -121,7 +121,7 @@ tasks:
   test:
     deps: [tools.gotestsum]
     cmds:
-      - GOWORK=off ./tools/gotestsum{{exeExt}} --format testname -- -v -coverprofile=coverage.txt ./internal/... ./template_funcs/... ./template/...
+      - GOWORK=off ./tools/gotestsum{{exeExt}} --format testname -- -v -coverprofile=coverage.txt ./config/... ./internal/... ./template_funcs/... ./template/...
     desc: run unit tests
     generates:
       - coverage.txt
--- a/config/config.go
+++ b/config/config.go
@@ -361,7 +361,11 @@ func (c *RootConfig) Initialize(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("getting package config: %w", err)
 			}
-			if recursivePkgConfig.Config.ShouldExcludeSubpkg(subpkg) {
+			excluded, err := recursivePkgConfig.Config.ShouldExcludeSubpkg(subpkg)
+			if err != nil {
+				return err
+			}
+			if excluded {
 				pkgLog.Debug().Msg("package was marked for exclusion")
 				continue
 			}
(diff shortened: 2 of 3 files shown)


LandonTClipp (approved):

LandonTClipp:
Looks good, thanks for the fix.

dylanpulver:
Heads up on the new commit, the Windows job was red on the previous head.

Adding `./config/...` to the test task runs `TestBuildTemplateData` in CI for the first time, and it asserted a hardcoded `"/config"`. `ConfigDir` is `filepath.Dir` output, which `Clean` rewrites to `\config` on Windows, so the expectation is spelled with `filepath.FromSlash` now. Production code is untouched.

`go test ./config/... ./internal/... ./template_funcs/... ./template/...` is clean locally. The run on the new head needs a maintainer approval to start.

PR #1179, opened 2026-08-25
v3: add cpuprofile option

### LINKED ISSUE #956: v3: Add `cpuprofile` option
We want to be able to take CPU profiles of mockery as can be done in v2.

ChrisJr404:
Description
-------------

Fixes #956. This brings the `cpuprofile` option back for v3, since it was dropped in the v2 to v3 move.

You can set it three ways, matching how other options work: `--cpuprofile <file>`, `MOCKERY_CPUPROFILE=<file>`, or `cpuprofile: <file>` in the config. When set, mockery writes a CPU profile of the run to that path so you can dig into slow generation. When empty (the default) nothing changes.

I also updated `migrate` to carry the old v2 `cpuprofile` value over instead of reporting it as unsupported.

## Type of change

- [x] 1.23

How Has This Been Tested?
---------------------------

Added unit tests for the config wiring (config file, flag, and default) and for the profiling helper (it produces a non-empty profile file, and errors on an unwritable path). Also ran `go test ./...` and generated a profile locally against a real package to confirm `go tool pprof` can open it.

Checklist
-----------

- [x] I have added tests that prove my feature works


LandonTClipp (approved):
This is great, thank you!

PR #1184, opened 2026-09-06
feat: support embedded types in matryer mocks

### LINKED ISSUE #1031: Support for embedding types into mocks in the `matryer` template
This issue is a feature proposal based on matryer/moq#226 (implemented in matryer/moq#227).

Some interfaces may have a mechanism to guarantee that a specific type is embedded.

For example, the `<ServiceName>Server` interface generated by `protoc-gen-go-grpc` contains an unexported `mustEmbedUnimplemented<ServiceName>Server` method, which can be implemented by embedding the `Unimplemented<ServiceName>Server` from the same package.

You can find more information here:
https://github.com/grpc/grpc-go/tree/master/cmd/protoc-gen-go-grpc#future-proofing-services

Real examples can be seen in the [helloworld example](https://pkg.go.dev/google.golang.org/grpc/examples/helloworld/helloworld) of grpc-go:

- https://github.com/grpc/grpc-go/blob/941102b7811f/examples/helloworld/helloworld/helloworld_grpc.pb.go#L72-L76
- https://github.com/grpc/grpc-go/blob/941102b7811f/examples/helloworld/helloworld/helloworld_grpc.pb.go#L83-L88

To support this pattern, I propose a feature that allows specified types to be embedded in mocks.

The `.mockery.yml` would likely look like this:

```yaml
packages:
  google.golang.org/grpc/examples/helloworld/helloworld:
    interfaces:
      GreeterServer:
        config:
          embed-types:
            - UnimplementedGreeterServer
```

Since the purpose is to embed types that implement unexported methods, the types used must belong to the same package as the interface.

eliotee:
## Description

Add `struct-preamble` and `add-import` template data to matryer mocks so interfaces requiring an embedded type, such as gRPC server interfaces with unexported marker methods, can be implemented outside their original package. Each mock can inject literal Go struct fields and request explicitly named imports; imports are shared across mocks in the same file, and conflicting names produce a generation error instead of silently changing the meaning of the preamble.

Document and validate the new options. Add compiled fixtures for external and in-package embedding, generic types, multiple interfaces, and explicit aliases. The generated fixtures account for most of the added lines. The testify template and existing import-registration behavior are unchanged.

Fixes #1031.

## Type of change

## Version of Go used when building/testing

Go 1.26.0 on macOS arm64.

## How Has This Been Tested?

- The original template generated an external mock that failed to compile because it did not implement the unexported interface method; the new embedding fixtures compile and pass.
- Schema, import-name conflict, deduplication, and generated mock behavior tests pass.
- `go run github.com/go-task/task/v3/cmd/task test test.e2e`: 199 unit tests, all shell e2e scenarios, and 5 Go e2e tests passed.
- `go run github.com/go-task/task/v3/cmd/task lint`: 0 issues.
- Repeated `mocks.generate` left all 243 Go files byte-for-byte identical; hand-written Go files were formatted with gofumpt, and `git diff --check` passed.

## Checklist

- [x] I have added tests that prove this feature works


template/registry.go
```diff
@@ -117,6 +118,36 @@ func (r *Registry) AddImport(pkgName string, pkgPath string) *Package {
 	})
 }
 
+// AddImportWithAlias добавляет импорт с точным именем, возвращая ошибку при конфликте.
+// Повторное добавление того же пути с тем же именем возвращает существующий импорт.
```
LandonTClipp: Could you convert this into English?

LandonTClipp (commented):
This is a great idea. We can get it merged soon but I will need to give it one more pass in the next few days. I am busy for the next two days.

template/registry.go
eliotee: Yep, sorry, my bad.

LandonTClipp (approved):
Looks good to me! Thanks for the good contribution.

PR #1185, opened 2026-09-10
fix(matryer): honor interface-level skip-ensure imports

### LINKED ISSUE #1082: v3: adds unused import in matryer template with template.skip-ensure
## Description

When source package, and target package name are different and template.skip-ensure is true, it adds an unused import

## Reproducer

```
all: false
template: matryer
force-file-write: true
packages:
    github.com/org/repo/package/subpackage:
        interfaces:
          Interfacer:
            config:
              pkgname: "packagemocks"
              dir: "./subpackage/packagemocks"
              filename: "mocks.go"
              structname: "InterfacerMock"
              template-data:
                skip-ensure: true
```

## Expected behavior

I would expect it wouldn't try to import an unused import.

## Mockery version

`v3.5.0`

## Installation Mechanism

- [X] `go get`
- [] Pre-built release
- [] homebrew
- [] Other: [please describe]

## Go version

`go version go1.24.4 darwin/arm64`

fzlzjerry:
Description
-------------

Fixes #1082.

Register the Matryer source-package import when at least one interface actually emits an ensure declaration. The import section previously checked file-level `TemplateData`, while the declarations check each interface's effective `TemplateData`.

That mismatch can produce either an unused source import (an interface opts out) or an undefined package qualifier (the file opts out and an interface opts back in).

This changes only the ensure-related registration. Configuration inheritance is unchanged; method-type imports remain intact, and explicit `add-import` entries are still processed before automatic ensure imports so a later interface can supply the source alias.

## Type of change

- [x] 1.26.8
- [x] 1.27.1

How Has This Been Tested?
---------------------------

- Twelve template/registry regression cases, including both override directions, mixed interfaces, same-package generation, existing type imports, and aliases supplied by later interfaces. Generated code is parsed and checked with `go/types`, not just formatted.
- Four end-to-end cases build and invoke the real CLI, compile the generated mocks in separate modules, and call the generated methods/check call histories. Three fail on the original implementation; all four pass with the fix. Four of the twelve template cases also fail before the fix.
- Eleven additional real CLI/consumer scenarios reproduce six original compiler failures; all eleven compile and run after the change. The five previously working controls retain byte-identical generated code.
- The repository's complete `test.ci` flow passes on Go 1.26.8 and 1.27.1: lint, complete mock regeneration, `git-state`, 217 unit-test events, shell generation checks, and 10 end-to-end test events. The Go 1.26 run was repeated with `--force` so Task did not reuse lint/tool-building cache decisions.
- `GOFLAGS='-mod=readonly -race' GOWORK=off go test -count=1 -json ./...` passes with 229 test events and no failures; the flag is inherited by the CLI/consumer subprocesses too.

Reproduce the focused tests:

```sh
GOWORK=off go test ./internal ./e2e/test_matryer_skip_ensure \
  -run '^(TestMatryerEnsureImports|TestInterfaceSkipEnsure)$' -count=1 -v
```

No generated fixtures, dependency manifests, or template-data merge rules are changed.

Checklist
-----------

No documentation change is needed: the existing interface-level option now generates compilable code consistently.

### DIFF
--- a/e2e/test_matryer_skip_ensure/skip_ensure_test.go
+++ b/e2e/test_matryer_skip_ensure/skip_ensure_test.go
@@ -1,0 +1,87 @@
+package test_matryer_skip_ensure
+
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestInterfaceSkipEnsure(t *testing.T) {
+	binary := filepath.Join(t.TempDir(), "mockery")
+	if runtime.GOOS == "windows" {
+		binary += ".exe"
+	}
+	build := exec.CommandContext(t.Context(), "go", "build", "-o", binary, "github.com/vektra/mockery/v3")
+	build.Env = append(os.Environ(), "GOWORK=off")
+	output, err := build.CombinedOutput()
+	require.NoError(t, err, string(output))
+
+	for _, tt := range []struct {
+		name         string
+		fileSkip     bool
+		firstSkip    bool
+		secondSkip   bool
+		mixed        bool
+		methodImport bool
+	}{
+		{name: "interface disables ensure", firstSkip: true},
+		{name: "interface enables ensure", fileSkip: true},
+		{name: "mixed shared output", fileSkip: true, mixed: true, secondSkip: true},
+		{name: "method retains source import", firstSkip: true, methodImport: true},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			require.NoError(t, os.Mkdir(filepath.Join(dir, "api"), 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(dir, "go.mod"), []byte("module example.com/skipensure\n\ngo 1.25.5\n"), 0o600))
+			source := "package api\ntype First interface { Ping() string }; type Second interface { Pong() string }; type Value struct { N int }; type Typed interface { Get() Value }\n"
+			require.NoError(t, os.WriteFile(filepath.Join(dir, "api", "api.go"), []byte(source), 0o600))
+			first := "First"
+			if tt.methodImport {
+				first = "Typed"
+			}
+			interfaces := map[string]any{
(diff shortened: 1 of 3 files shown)


e2e/test_matryer_skip_ensure/skip_ensure_test.go
```diff
@@ -0,0 +1,87 @@
...
+		{name: "interface enables ensure", fileSkip: true},
+		{name: "mixed shared output", fileSkip: true, mixed: true, secondSkip: true},
+		{name: "method retains source import", firstSkip: true, methodImport: true},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			require.NoError(t, os.Mkdir(filepath.Join(dir, "api"), 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(dir, "go.mod"), []byte("module example.com/skipensure\n\ngo 1.25.5\n"), 0o600))
+			source := "package api\ntype First interface { Ping() string }; type Second interface { Pong() string }; type Value struct { N int }; type Typed interface { Get() Value }\n"
+			require.NoError(t, os.WriteFile(filepath.Join(dir, "api", "api.go"), []byte(source), 0o600))
```
LandonTClipp, line 42: Instead of writing out the test file each iteration, can you make a test fixture with this source file, add the config to the root mockery config file, and then write a test file that asserts the behavior of the mock file? Adding mock generation in each test like this will quickly balloon the test run time. That's why the tests generate all of the mock files once and then assert the result of those outputs.

LandonTClipp (changes requested):

e2e/test_matryer_skip_ensure/skip_ensure_test.go
fzlzjerry: Done in 75f59a6. I removed the e2e test and added fixture packages under `internal/fixtures/matryer_skip_ensure` (`ensure` keeps the default; `skip` sets `skip-ensure: True` at package level), with their config in `.mockery_matryer.yml`. The four scenarios are generated once with the rest of the mocks into a separate `mocks` package, and `skip_ensure_test.go` asserts on the committed output: interface conformance, call histories, and whether the ensure declarations and source imports appear as configured. With the template fix reverted, the regenerated package fails to compile (`undefined: skip` twice and one unused `ensure` import).

